PR #407, opened 2017-04-01
Travis and Appveyor

fschulze:
Currently pypy doesn't work, because argon2 doesn't install correctly and the client tests hang with Appveyor. This is still very useful to see if basic stuff breaks in PRs etc. For releases we still need the old ways of testing.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -6,10 +6,10 @@ environment:
     - PROJECT: "client"
 
 install:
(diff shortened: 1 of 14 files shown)


hpk42:
can we skip pypy tests on windows for now?  I don't think it's relevant and it's better to have a green default state ...

fschulze:
I already skip pypy on Windows, the failure is due to hanging devpi-client tests which I still have to investigate.

fschulze:
And we are all green!

PR #411, opened 2017-04-09
Add option to `devpi install` to use requirements file

andrewleech:
Adds -r, --requirement option to `devpi install`.
A standard requirements.txt file is then expected for the pkg argument.
These args will be passed through to pip for the install.
Includes basic unit test.

### DIFF
--- a/client/devpi/main.py
+++ b/client/devpi/main.py
@@ -855,6 +855,8 @@ def install(parser):
     parser.add_argument("--venv", action="store", metavar="DIR",
         help="install into specified virtualenv (created on the fly "
              "if none exists).")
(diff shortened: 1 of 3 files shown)


client/devpi/main.py
```diff
@@ -855,6 +855,8 @@ def install(parser):
     parser.add_argument("--venv", action="store", metavar="DIR",
         help="install into specified virtualenv (created on the fly "
              "if none exists).")
+    parser.add_argument("-r", "--requirement", action="store_true",
```
RonnyPfannschmidt: i would like to suggest to change this to a collecting list, so people dont get surprised if they use -r more than once or mix requirement files and normal requirements

PR #412, opened 2017-04-09
Use activated venv

andrewleech:
This is a combination/extension of my virtualenv pull requests from bitbucket.

It also includes automatic detection of running from virtualenv and/or running with activated virtualenv and using this for `use`, `install` etc.
As such the previous runtime argument --venv=yes is removed as this is implied now.

The changeset does include the venv handling in --set-cfg in that when operating in a venv only the pip.conf of the venv will be set, not the global settings. There is a line in the output notifying the user as such.

I haven't explicitly tested it yet but I expect it to work with upload as discussed in issue #395

I haven't yet added proper test coverage for the automatic virtualenv handing, I need to look into how venv is currently set up in similar tests.

### DIFF
--- a/client/devpi/use.py
+++ b/client/devpi/use.py
@@ -343,6 +343,18 @@ def getvenv():
         return None
     return pip.dirpath().dirpath()
 
+def current_venv():
+    venv = None
+    if (hasattr(sys, 'real_prefix') or
+            (hasattr(sys, 'base_prefix') and sys.base_prefix != sys.prefix)):
+        venv = sys.prefix
+
+    elif "VIRTUAL_ENV" in os.environ:
+        venv = os.environ["VIRTUAL_ENV"]
+
+    return venv
+
+
 def main(hub, args=None):
     args = hub.args
     current = hub.current
@@ -366,7 +378,12 @@ def main(hub, args=None):
 
     if args.venv:
         if args.venv != "-":
-            venvname = args.venv
+            if args.venv == "yes":
+                venvname = current_venv()
+                if venvname is None:
+                    hub.fatal('Could not detect an activated virtualenv.')
+            else:
+                venvname = args.venv
             cand = hub.cwd.join(venvname, vbin, abs=True)
             if not cand.check():
                 cand = hub.path_venvbase.join(venvname, vbin)
--- a/client/devpi/use.py
+++ b/client/devpi/use.py
@@ -376,26 +376,26 @@ def main(hub, args=None):
     if url or current.index:
         current.configure_fromurl(hub, url, client_cert=args.client_cert)
 
-    if args.venv:
-        if args.venv != "-":
-            if args.venv == "yes":
-                venvname = current_venv()
-                if venvname is None:
-                    hub.fatal('Could not detect an activated virtualenv.')
-            else:
-                venvname = args.venv
+    if args.venv == "-":
+        current.reconfigure(dict(venvdir=None))
+    else:
+        if args.venv:
+            venvname = args.venv
+        else:
+            venvname = current_venv()
+
+        if venvname:
             cand = hub.cwd.join(venvname, vbin, abs=True)
             if not cand.check():
                 cand = hub.path_venvbase.join(venvname, vbin)
                 if not cand.check():
                     hub.fatal("no virtualenv %r found" % venvname)
             current.reconfigure(dict(venvdir=cand.dirpath().strpath))
-        else:
-            current.reconfigure(dict(venvdir=None))
+
     if args.list:
-        if not hub.current.rooturl:
+        if not current.rooturl:
             hub.fatal("not connected to any server")
-        r = hub.http_api("GET", hub.current.rooturl, {}, quiet=True)
+        r = hub.http_api("GET", current.rooturl, {}, quiet=True)
         out_index_list(hub, r.result)
         return 0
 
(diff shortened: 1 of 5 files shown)


client/devpi/use.py
```diff
@@ -343,6 +343,18 @@ def getvenv():
         return None
     return pip.dirpath().dirpath()
 
+def current_venv():
```
RonnyPfannschmidt: the order here could use a review, imagine devpi being made in one virtualenv and the user working in another

andrewleech: Yes I had thought of that use case earlier on and wondered about it, forgot to take it into consideration here. I think you're right, it's probably better to take the env var as more important than the current python runtime.

andrewleech: I did switch the order around in a commit since this, I do think it's much better now.

andrewleech:
I'd like to think this could be good to go, any chance of a review?

fschulze:
I miss updated command line option descriptions, that's why I thought this isn't done yet.

pi-anl:
Ah, I don't remember what I did there, but yes you're probably right. I'll check it again when I get a chance!

fschulze:
1. Why is the ``pip.conf`` created in ``<virtualenv>/bin/``? Shouldn't it be directly in the virtualenv directory? With ``--venv=.`` it creates it directly in the virtualenv directory, not in ``bin``.
2. After trying it, I'm not convinced using the ``sys`` prefix stuff by default makes sense. I wonder how many people use tools like ``pipsi`` or install tools in separate venvs to allow version pinning etc. For all of those people ``--set-cfg`` would go into some obscure place. Maybe it should only use that location if you are currently in the same path? Then something like ``./bin/devpi use ...`` would still do the right thing even without an activated venv.
3. Using a non existing path for ``--venv`` causes a traceback.
4. There currently is no way to see which configuration files are changed before it's done. I thought about a confirmation prompt, but am not sure whether that would disrupt regular use. Not sure this is a blocker for this PR.
5. I never use ``--always-set-cfg``, so I'm unsure how that plays into this PR.

andrewleech:
1.  I don't see it installing into `<virtualenv>/bin/` myself, my testing always has it go into `<virtualenv>`.  Have you seen it make the file in the wrong location, if so what was you command line?

2. Yeah I've been thinking about this one myself and I agree it could break an unknown number of people's standard use case. I'm not sure about it being `cd` based, that feels a little too 'black magic' to me, it's likely to confuse users too. Personally when I'm running a tool like pip directly from .venv I rarely cd to venv first, I guess if I was going to do that I would be just as likely to activate the venv.
I was thinking perhaps a devpi (global) config option to enable devpi recognising the venv it's installed into. Then if you're the sort of person that installs devpi into every venv you make you only need to set the config once (like `--always-set-cfg `) and it'll work for you. This should probably be a separate commit at this point though.

3. I'll look into that, sounds like another unit test needs writing

4. Yeah that's a good point. When you just call `devpi use` it should possibly print this out. Alongside the current behaviour of printing the current index, printing the currently used config files would likely be really useful.

5. I'm pretty sure `--always-set-cfg` behaves consistently, at least the way I would expect it to. If it's previously been enabled it's basically like adding `--set-cfg` to every command line, so devpi will set the global/venv config depending on current runtime env.

fschulze:
1. It happens whenever the result of ``active_venv`` is used, which joins the path with ``vbin``. See https://github.com/andrewleech/devpi/blob/c2e4442f8fa29f73453b5bf099098d9ff9e427fa/client/devpi/use.py#L355
2. Maybe for now ``active_venv`` should only use the ``VIRTUAL_ENV`` environment variable. On the other hand, pip works the way your PR does now. So for ``devpi install`` the current PR behaviour would make sense. It's only ``--set-cfg`` which I'm unsure about.

Speaking of ``devpi install`` ... is that affected at all in this PR? If not, it might make more sense to make the changes there instead of messing with config file locations?

Usability and expectations due to other tools are hard ...

andrewleech:
For now I've removed the functionality to detect & use the virtualenv that devpi is installed in, doing that by default could break a number of people's existing installations.

It warrants further thought though, in some ways this might be a better default so as to match pip functionality for instance. However I think it needs to be configurable either way to that users can still install devpi in a venv for use as a standalone tool.
Worth raising a separate issue about I'd say.

I've also consolidated the venv detection and creation mechanism and the same is used for both `devpi install` and `devpi use`, so usage should now be consistent for both.

The crash you found is also fixed now, with improved unit test coverage to suit.

fschulze:
The ``create_virtualenv`` is new, so you should update the help text of ``--venv`` accordingly.

I will test it for a bit on my machine to see how it feels.

andrewleech:
Yeah good point, the on-the-demand creation was always there for `devpi install` but did not behave that way for `devpi use`. I've updated the help text as such.

fschulze:
Sorry for the long delay on this.

I like the changes. I made a rebased branch at https://github.com/fschulze/devpi/tree/use_activated_venv

I bumped the major version, because this is a backward incompatible change, even though it only affects people with activated virtualenvs. Any objections on that?

client/devpi/use.py
```diff
@@ -403,35 +398,42 @@ def main(hub, args=None):
...
-                PipCfg().clear_trustedhost(indexserver)
+                pipcfg.clear_trustedhost(indexserver)
 
-    show_one_conf(hub, DistutilsCfg())
-    show_one_conf(hub, PipCfg())
-    show_one_conf(hub, BuildoutCfg())
-    hub.line("always-set-cfg: %s" % ("yes" if hub.current.always_setcfg else
-                                     "no"))
+            set_cfgs.append(pipcfg)
+    for cfg in set_cfgs:
```
fschulze: I just noticed, that the current config settings aren't shown anymore. Only when they are changed. I think that should be fixed.
Before one got output like this from ``devpi use``:
```
$ devpi use /root/pypi
current devpi index: http://localhost:3141/root/pypi (logged in as emilie)
~/.pydistutils.cfg     : http://localhost:4040/alice/dev/+simple/
~/.pip/pip.conf        : http://localhost:4040/alice/dev/+simple/
~/.buildout/default.cfg: http://localhost:4040/alice/dev/+simple/
always-set-cfg: no
```
Now the three lines starting with ``~`` are missing.

andrewleech: Sure, that's certainly an unintentional change, I'll try to look into it soon.

andrewleech:
@fschulze  The config printouts look good again.

Current behaviour is:
* With no virtualenv declared/active: print all three system config paths (pip buildout, distutils)
* With virtualenv declared/active: print out just the pip.ini in the virtualenv itself

PR #416, opened 2017-04-11
Cleanup after dropping Python 2.6

fschulze:


hpk42:
there is a minor superflous "import sys" in client, failing the tests. otherwise it's all good.

PR #422, opened 2017-05-04
#345 - pip caching

fschulze:
See #345 

This also adds towncrier setup for devpi-server.

### DIFF
--- a/server/devpi_server/cfg/nginx-devpi.conf.template
+++ b/server/devpi_server/cfg/nginx-devpi.conf.template
@@ -36,6 +36,5 @@ server {
         proxy_pass http://localhost:%(port)s;
         proxy_set_header X-outside-url $scheme://$host:$server_port;
         proxy_set_header X-Real-IP $remote_addr;
-        expires -1;  # no-cache
     }
 } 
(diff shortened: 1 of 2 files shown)


StephanErb:
I fear that this could interact with the commonly deployed caching in Nginx. 

For example, we have to cache simple pages for 60 seconds as Devpi will collapse under the load otherwise. With this change we would now have to strip the newly introduced headers some how.

fschulze:
If you set ``Cache-Control`` headers in nginx and maybe remove the ``Expires`` header, then all should be good. Only the ``Expires`` and ``Cache-Control`` headers are added by devpi-server now. The default nginx.conf generated by devpi-server contains ``expires -1;`` which already overwrites both headers (and should actually be removed in this PR).

Hmm, I did some more tests now and it seems like the ``expires -1;`` is the actual problem. If I remove it, the simple pages have no ``Expires`` and ``Cache-Control`` headers and the caching lib used by pip doesn't cache the page anymore even with older devpi-server setups. Same with going directly to devpi-server.

So, I have to redo this PR and only change the default nginx config.

PR #425, opened 2017-05-05
Decode URL encoded paths while saving packages

vartec:
As of now devpi server does not decode the path component of the URL before using it as file path. 

This causes problems with characters that are valid in package name, but do get quoted (`!` and `+` are the ones I stumbled upon, both are valid characters in version specifier per PEP440).

Before the fix:

```reading remote: https://host/foo-1.2.3%2Bvendor-py2-none-any.whl, target root/a_mirror/+e/host/foo-1.2.3%2Bvendor-py2-none-any.whl```

Saving as `foo-1.2.3%2Bvendor-py2-none-any.whl` was subsequently causing `file not found` error when trying to access `foo-1.2.3+vendor-py2-none-any.whl`

After the fix:
```reading remote: https://host/foo-1.2.3%2Bvendor-py2-none-any.whl, target root/a_mirror/+e/host/foo-1.2.3+vendor-py2-none-any.whl```

### DIFF
--- a/client/CHANGELOG
+++ b/client/CHANGELOG
@@ -4,6 +4,7 @@
 3.0.1 (Unreleased)
 ==================
 
+- fix issue227: add ``logout`` command as alias for ``logoff``.
 
 
 3.0.0 (2017-04-23)
--- a/client/devpi/main.py
+++ b/client/devpi/main.py
@@ -679,6 +679,15 @@ def logoff(parser):
     This will erase the client-side login token (see "devpi login").
     """
 
+
+@subcommand("devpi.login:logoff")
+def logout(parser):
+    """ log out of the current devpi-server.
+
+    This will erase the client-side login token (see "devpi login").
+    """
+
+
 @subcommand("devpi.index")
 def index(parser):
     """ create, delete and manage indexes.
--- a/client/testing/test_functional.py
+++ b/client/testing/test_functional.py
@@ -68,6 +68,9 @@ class Mapp(MappMixin):
     def logoff(self, code=None):
         self.devpi("logoff", code=code)
 
+    def logout(self, code=None):
+        self.devpi("logout", code=code)
+
     def use(self, indexname):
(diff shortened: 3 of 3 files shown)


fschulze:
When exactly does this happen? I have devpi instances which have source releases with ``+`` in them and they work just fine. I used ``devpi upload`` to create them and pip happily uses them.

vartec:
This happens when you have a mirror with such a filenames.

fschulze:
I guess the changes in devpi-common are for already mirrored files? Otherwise the changes in devpi-server seem to be enough.

Could you try to write tests? See ``TestMirrorIndexThings`` in ``functional.py`` for other mirroring related tests.

vartec:
Sure, I'll give it a go on Monday.

fschulze:
Looks good!

Could you please add a changelog entry? See https://github.com/devpi/devpi/tree/master/server/news and https://github.com/devpi/devpi/blob/master/CONTRIBUTING.rst#pull-requests

Would be great if you could give some feedback on the ``CONTRIBUTING.rst`` document. Would you add something, is the wording ok?

Am I correct in my assumption that the changes in devpi-common are only needed for already mirrored files?

vartec:
> Am I correct in my assumption that the changes in devpi-common are only needed for already mirrored files?

It's for the `if link.hash_spec` code path in `maplink`. Does that mean already mirrored? (I admit, I didn't dig super deep into the code)

fschulze:
> > Am I correct in my assumption that the changes in devpi-common are 
> only needed for already mirrored files?
>
> It's for the `if link.hash_spec` code path in `maplink`. Does that 
> mean already mirrored? (I admit, I didn't dig super deep into the 
> code)

I'll look into it before merging. Thanks for the pointer.

vartec:
oh..bummer, I've messed up the rebase, give me a while to clean that up..

PR #429, opened 2017-05-06
Fix #386: Add ``passwd`` command to enable password setting with interactive input.

### LINKED ISSUE #386: prompt for password instead of passing it on the commandline
- Bitbucket: https://bitbucket.org/hpk42/devpi/issue/386
- Originally reported by: [jeroen1205](http://bitbucket.org/jeroen1205)
- Originally created at: 2017-01-20T16:16:26.850

When I change the password of the devpi root user, I need to specify it on the commandline:

```
devpi user -m root password=mypassword
```

This is insecure, as this might put the password in plaintext in your bash_history (or similar shell history file).

Is there an option to let devpi prompt for the password (and don't echo it to the screen ofc.)?

fschulze:


client/devpi/user.py
```diff
@@ -21,6 +27,8 @@ def user_modify(hub, user, kvdict):
     reply = hub.http_api("get", url, type="userconfig")
     for name, val in kvdict.items():
         reply.result[name] = val
+        if name == 'password':
+            val = '********'
```
hpk42: i don't understand this line -- setting to '*******' -- what does it mean?

client/devpi/main.py
```diff
@@ -659,6 +659,14 @@ def user(parser):
         help="key=value configuration item.  Possible keys: "
              "email, password.")
 
+
+@subcommand("devpi.user:passwd")
+def passwd(parser):
+    """ change user password. """
+    parser.add_argument(
+        "username", type=str, action="store", nargs="?", help="user name")
```
hpk42: the help string is not very telling -- i guess it's "set password for the specified ..."

client/devpi/user.py
fschulze: It prevents the output of the password in the log. I added a comment.

client/devpi/main.py
fschulze: I changed the wording now.

PR #430, opened 2017-05-08
Decode URL encoded paths while saving packages

vartec:
As of now devpi server does not decode the path component of the URL before using it as file path.

This causes problems with characters that are valid in package name, but do get quoted (! and + are the ones I stumbled upon, both are valid characters in version specifier per PEP440).

Before the fix:

```reading remote: https://host/foo-1.2.3%2Bvendor-py2-none-any.whl, target root/a_mirror/+e/host/foo-1.2.3%2Bvendor-py2-none-any.whl```

Saving as `foo-1.2.3%2Bvendor-py2-none-any.whl` was subsequently causing file not found error when trying to access `foo-1.2.3+vendor-py2-none-any.whl`

After the fix:
```reading remote: https://host/foo-1.2.3%2Bvendor-py2-none-any.whl, target root/a_mirror/+e/host/foo-1.2.3+vendor-py2-none-any.whl```

### DIFF
--- a/common/devpi_common/url.py
+++ b/common/devpi_common/url.py
@@ -1,13 +1,14 @@
-
 import sys
 import posixpath
 from devpi_common.types import cached_property, ensure_unicode, parse_hash_spec
 from requests.models import parse_url
 
 if sys.version_info >= (3, 0):
-    from urllib.parse import urlparse, urlunsplit, urljoin
+    from urllib.parse import urlparse, urlunsplit, urljoin, unquote
 else:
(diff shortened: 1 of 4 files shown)


vartec:
This is same thing that was supposed to be in https://github.com/devpi/devpi/pull/425, but I messed up rebasing

fschulze:
I added a test for the devpi-common part and in so doing found and fixed another bug. Thanks for the PR!

PR #431, opened 2017-05-09
Postgresql tests via plugin

fschulze:


postgresql/tox.ini
```diff
@@ -14,6 +14,14 @@ deps =
     pytest-catchlog
     pytest-flakes
     pytest-pep8
+    -e
+    {toxinidir}/../common
+    -e
+    {toxinidir}/../server
```
hpk42: putting "-e" and the location in different lines is intentional?   I thought it needs to be on the same line.

server/devpi_server_pytest/__init__.py
```diff
@@ -0,0 +1,7 @@
+def pytest_addoption(parser):
+    parser.addoption(
+        "--slow", action="store_true", default=False,
+        help="run slow tests involving remote services (pypi.python.org)")
+    parser.addoption(
+        "--backend", action="store",
+        help="run tests with specified dotted name backend")
```
hpk42: hum, this introduces a "completion" conflict when i type "vim devpi_serve<TAB>" which currently does not exist.   You could also put the plugin into the test_devpi_server directory  which kind of makes sense i'd think.

postgresql/tox.ini
fschulze: Yes, it seems this is a bug in tox, it doesn't work when it's on one line.

server/devpi_server_pytest/__init__.py
fschulze: I've moved the plugin. I had problems before, which I thought where caused by placing it in test_devpi_server, but they were unrelated.

fschulze:
Moving the plugin into test_devpi_server doesn't work after all: https://travis-ci.org/fschulze/devpi/jobs/230385547

fschulze:
Is there a way to fix the issue? Or should I move the plugin to ``pytest_devpi_server`` to avoid the auto complete conflict?

postgresql/tox.ini
obestwalter: Not a bug - a documented limitation: "Each line specifies a file, a URL or a package name." [*](http://tox.readthedocs.io/en/latest/config.html#confval-deps=MULTI-LINE-LIST) (so, no options) - there is no way to add options as each line is taken as is and not split before passed to pip. If you need options you either need to use a requirements file or stick you pip invocation directly into commands.

But you worked around this limitation beautifully :)

fschulze: Options could be supported by doing (pseudocode):
```python
if option.startswith('-'):
    parts = option.split(None, 1)
    args.extend(parts)
```
or am I missing something?

obestwalter: I agree that it seems fixable with a bit of extra effort. Question is, if someone wants it strong enough to add it. I never missed it and was not even aware of that limitation until I saw this :)

fschulze: True

obestwalter: btw sorry for necromancing this - I just realized that the last comment here was months ago :D

PR #438, opened 2017-06-29
Add "devpiserver_on_remove" hook for removing unpacked doczip files

drtuxwang:
Currently "devpi_web" unpacks "doc.zip" but these unpacked directories are never deleted when a ".doc.zip" file is removed from the devpi registry. This changes adds a new hook which calls into removal code in "devpi_web"


RonnyPfannschmidt:
looks good, can you add a test or do you need help with figuring the details

drtuxwang:
Looks like there is a "test_doczip_uploaded_hook" test in "server/test_devpi_server/test_model.py" but I cannot find any test relating to this in "web/devpi_web".

I can write a "test_doczip_remove_hook" test using this test as a guide. Unfortunately I am unable to work out how to run the tests.

drtuxwang:
Ronny,

Hello, I have added a test. But have problems working out how to run it.

Can you take look please?

Thanks

Colin

fschulze:
I took a look at this. Compared to the other hooks, this new one is in the wrong place. The other hooks trigger from the event thread, the new one would trigger from the main thread. I tried moving it, but didn't get it working yet, because in the correct place I don't have enough data for a useful hook. I will investigate this further.

drtuxwang:
Hi, any news about merging this into master?

Thanks

Colin

drtuxwang:
Hi I have successfully moved remove docs call to the TxNotificationThread and tested it. Can you take another look please? Thanks

server/test_devpi_server/test_model.py
```diff
@@ -780,6 +780,36 @@ def devpiserver_on_upload(self, stage, project, version, link):
             assert link.entry.file_exists()
 
 
+    @pytest.mark.start_threads
+    def test_doczip_remove_hook(self, stage, queue):
+        class Plugin:
+            def devpiserver_on_remove(self, stage, link):
```
fschulze: The mock plugin needs to be updated for the new hook signature.

drtuxwang:
Hi, I have fixed the unit test and added a line to the ./server/CHANGELOG

However, I am unable to work out why tox has problems finding devpiserver_on_remove. Can you give me advice please? Thanks

fschulze:
I've fixed that issue and merged your changes after a rebase.

drtuxwang:
Thanks :)

PR #453, opened 2017-08-11
Fix replication tests

fschulze:
This PR fixes the devpi-server tests.

### DIFF
--- a/client/testing/test_functional.py
+++ b/client/testing/test_functional.py
@@ -14,6 +14,14 @@ except ImportError:
         def test_mirror_things(self):
             pytest.skip(
                 "Couldn't import TestMirrorIndexThings from devpi server tests.")
+try:
+    from test_devpi_server.functional import TestIndexPushThings  # noqa
+except ImportError:
+    # when testing with older devpi-server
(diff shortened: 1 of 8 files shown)


client/testing/test_functional.py
```diff
@@ -14,6 +14,14 @@ class TestMirrorIndexThings:
...
             pytest.skip(
                 "Couldn't import TestMirrorIndexThings from devpi server tests.")
+try:
+    from test_devpi_server.functional import TestIndexPushThings  # noqa
+except ImportError:
+    # when testing with older devpi-server
+    class TestIndexPushThings:
+        def test_mirror_things(self):
+            pytest.skip(
+                "Couldn't import TestIndexPushThings from devpi server tests.")
```
hpk42: can we not test for the version of devpi-server instead of try-importing?  If importing fails
for other than version-reasons we'd quite silently skip the tests here which is kind of bad practise ...

fschulze: We could, but in this case the import would fail in line 8 already and doing it by version is more complex and during development more fragile. So I don't see the point. I agree in general about abusing ImportError though.

PR #461, opened 2017-09-04
HTML doesn't have a ``PUSH`` method, replace it with ``POST``.

fschulze:
For backward compatibility we still need to support ``PUSH`` though. But this fixes warnings in tests.


hpk42:
can there not be arbitrary verbs for http resources?  I am still merging this because it's more a theoretical question and doesn't much matter in practise i guess.

PR #472, opened 2017-11-08
devpi-server: Do not install tests

seifertm:
I am currently packaging devpi-server for Gentoo Linux and noticed that setup.py installs the tests into site-packages.

This PR simply excludes the test folder from find_packages, so that they will not be installed.

### DIFF
--- a/server/setup.py
+++ b/server/setup.py
@@ -43,7 +43,7 @@ if __name__ == "__main__":
       version='4.3.1rc1',
       maintainer="Holger Krekel, Florian Schulze",
       maintainer_email="holger@merlinux.eu",
-      packages=find_packages(),
+      packages=find_packages(exclude=["test_*"]),
       include_package_data=True,
       zip_safe=False,
       license="MIT",


fschulze:
The other packages currently import from there for testing. Is there a way to install the tests only with an "extras" like ``devpi-server[tests]``? Then we could use that.

seifertm:
Ahh, I see. Didn't know the tests were connected to the other projects.

Wouldn't it make sense then to move the tests to a separate project, since they are a jointly used test suite? Then they could have their own setup.py and packages could depend on them independently from the production packages.

Another possibility would be to implement a custom install, such as ``python setup.py install_tests``. I didn't find a solution that uses extras to install packages separately, as you proposed.

However, this turns out to be a bigger change than intended. It certainly would be nice to have devpi-server install without tests, but it's not a deal breaker :)
Do you think the benefits outweigh the costs here?

fschulze:
It would be quite a bit of work to separate the tests. We want ``devpi test devpi-server`` to work for example. The test package shouldn't conflict with any other package, so I don't really see harm in it being installed.

seifertm:
If it involves a lot of work to change it, then let's just close this PR.

Thanks for the help!

PR #478, opened 2017-11-23
Fix issue477: importing with --no-root-pypi failed

fschulze:


server/devpi_server/model.py
```diff
@@ -134,10 +134,14 @@ def getstage(self, user, index=None):
 
     def is_empty(self):
```
gotcha: Might be useful to rename this method to `may_import`.
It seems to be used only by `./server/devpi_server/importexport.py`

server/devpi_server/model.py
```diff
@@ -134,10 +134,14 @@ def getstage(self, user, index=None):
...
         userlist = self.get_userlist()
+        if len(userlist) == 0:
+            return True
         if len(userlist) == 1:
             user, = userlist
             if user.name == "root":
-                rootindexes = user.get().get("indexes", [])
+                rootindexes = list(user.get().get("indexes", []))
```
gotcha: Adding a comment here to relate it to `--no-root-pypi`

fschulze:
@gotcha made changes as requested

server/devpi_server/importexport.py
```diff
@@ -35,8 +53,8 @@ def do_import(path, xom):
     if not path.check():
         fatal("path for importing not found: %s" %(path))
 
-    with xom.keyfs.transaction():
-        if not xom.model.is_empty():
+    with xom.keyfs.transaction(write=False):
+        if not no_existing_data(xom):
```
gotcha: Nice move to use a function.
I would name it `has_users_or_stages` (referring to the error message in the exception) and thus avoid the double negation.

server/devpi_server/importexport.py
```diff
@@ -35,8 +54,8 @@ def do_import(path, xom):
     if not path.check():
         fatal("path for importing not found: %s" %(path))
 
-    with xom.keyfs.transaction():
-        if not xom.model.is_empty():
+    with xom.keyfs.transaction(write=False):
```
gotcha: For my information, why is there `write=False` ?

fschulze: Just wanted to make it explicit that it's a read only transaction. It wouldn't be strictly necessary, because it's the default anyway.

gotcha:
LGTM

PR #481, opened 2017-12-08
480 additional metadata

### LINKED ISSUE #480: push to pypi doesnt include metadata, casues broken pytest metadata on pypy
followup to pytest-dev/pytest#2966

the pytest release that was pushed by devpi doesnt propperly propagate the requires_python metadata to pypi

fschulze:
This should fix #480 and further instances of https://github.com/pytest-dev/pytest/issues/2966 by storing the additional metadata.

I tested "push" from one devpi to another and the new metadata is transferred.

Uploads with "devpi upload" and "twine upload" also store the metadata properly now with no additional fixes.

@jaraco @nicoddemus @RonnyPfannschmidt @takluyver

If this looks good, I will update devpi.net to use a prerelease for further testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ python:
   - "3.6"
   - "nightly"
   - "pypy-5.4.1"
-  - "pypy3.3-5.2-alpha1"
+#  - "pypy3.3-5.2-alpha1"
 env:
   - PROJECT=client
   - PROJECT=common
--- a/postgresql/tox.ini
+++ b/postgresql/tox.ini
@@ -11,7 +11,6 @@ deps =
     webtest
     mock
     pytest
-    pytest-catchlog
     pytest-flakes
     pytest-pep8
     -e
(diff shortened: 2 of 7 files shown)


fschulze:
To be clear, I'm not asking for devpi internal review, but whether the list of metadata keys looks sane. Maybe someone also has a better idea where to get the keys from, so we wouldn't have to maintain the list ourself.

.travis.yml
```diff
@@ -9,7 +9,7 @@ python:
   - "3.6"
   - "nightly"
   - "pypy-5.4.1"
-  - "pypy3.3-5.2-alpha1"
+#  - "pypy3.3-5.2-alpha1"
```
hpk42: just strike it.  it's an odd version to test against, anyway.

hpk42:
this looks good. I don't think this changes much -- such changes take their time to proliferate anyway. so i think it's fine to just maintain this list and fix it when an issue occurs. (my 2cent)

PR #484, opened 2017-12-11
fix issue482: let external links in documentation open outside of iframe.

fschulze:


gotcha:
LGTM (did it together)

PR #485, opened 2017-12-11
Versioned static resources

fschulze:


gotcha:
LGTM when tests pass

RonnyPfannschmidt:
how does this behave with a webserver in front? (thats possibly serving the static files)

fschulze:
@RonnyPfannschmidt With the example nginx config only the releases and docs are served by nginx and they have a different route. I only change the URL for the static files route in the devpi-web package and for the theme path option.

RonnyPfannschmidt:
i see, thanks for clarifying 👍

PR #486, opened 2017-12-11
469 acl toxresult upload

gotcha:


fschulze:
See https://github.com/gotcha/devpi/pull/1 for docs

Issue #489, opened 2017-12-12
Replication error


fschulze:
A more complex situation:
- a release was mirrored from pypi
- at some point the locally cached file was removed from the filesystem
- the release/project was removed from pypi
- a new replica tries to fetch the file, because it's still mentioned in the changelog
- the master tries to download it from pypi and gets a 404
- the master replies with a 502 and the replica stops at this point

fschulze:
Fixed with devpi-server 4.3.2.

PR #490, opened 2017-12-14
489 replication error

fschulze:


fschulze:
Fixes #489

fschulze:
@hpk42 I'm not entirely sure if it's ok to open a read transaction in ImportFileReplica since that is called by the keyfs event handler

server/devpi_server/replica.py
```diff
@@ -427,8 +427,19 @@ def __call__(self, fswriter, key, val, back_serial):
                            relpath)
             return
 
+        if r.status_code in (404, 502):
+            stagename = '/'.join(relpath.split('/')[:2])
+            with key.keyfs.transaction(write=False):
+                stage = self.xom.model.getstage(stagename)
+            if stage.ixconfig['type'] == 'mirror':
```
hpk42: i'd move the "if stage.ixconfig" into the transaction.

server/devpi_server/replica.py
```diff
@@ -427,8 +427,19 @@ def __call__(self, fswriter, key, val, back_serial):
                            relpath)
             return
 
+        if r.status_code in (404, 502):
+            stagename = '/'.join(relpath.split('/')[:2])
+            with key.keyfs.transaction(write=False):
```
hpk42: maybe this needs a specific serial if it's runnong out of sync.  The stage might have been deleted in newer serials.

hpk42: hum, wait a second. isn't the whole method executing as part of a write-transaction (see ``KeyFS.import_changes``)?

hpk42: if it's running as part of importing changes into a replica, then we should automatically be at the "newest" and correct serial for this import.  But it shouldn't need a transaction ... is my current reading.  can't currently look much deeper.

fschulze: I looked into it some more. That part is pretty confusing, because we deal with two different storages/transactions. One is the underlying storage sqlite/postgresql and the other is our own keyfs. The transaction in import_changes was the lower level sqlite/postgresql. I now added opening a read transaction outside the subscriber call to make it explicit and avoid having to push in the current serial. With the current design the subscriber is always called as soon as the serial was imported, so there can't be newer transactions where the index would have been deleted, that is why the replication stopped in the first place. I added removal of the index to the test in case we change something in the future.

hpk42:
thanks, LGTM now!

PR #493, opened 2017-12-18
Use Travis build stages to first upload packages to devpi.net and then use them when running tests.

fschulze:
See:
- https://docs.travis-ci.com/user/build-stages/
- https://devpi.net/devpi-travis/fschulze_devpi_travis-stages

This allows us to test all packages with latest dependencies. So devpi-web etc is tested against development devpi-server instead of only the last release.

This also avoids the explicit "-e ../server" hack for dependencies in tox.ini I used before.

We could also add a stage for building docs now.


fschulze:
Hrm, one huge problem. PR builds don't work. The secret on travis isn't used for security reasons. Anyone could exfiltrate the secret otherwise. I guess we have to detect that case and run with normal releases. Or we have to go with my first idea and run a local devpi-server for each travis build.

fschulze:
For now I disabled use of devpi.net for pull requests.

.travis.yml
```diff
@@ -11,10 +11,50 @@ python:
...
+  include:
+    - stage: create packages
+      addons: []
+      install: pip install devpi-client towncrier
+      python: "2.7"
+      env: []
+      script:
+        - devpi use https://devpi.net/devpi-travis
+        - devpi login devpi-travis --password ${DEVPI_TRAVIS_PASSWD}
+        - yes yes | devpi index --delete "${DEVPI_INDEX}" || true
```
hpk42: looks like we could add a ``--yes-really`` option sometime.

fschulze:
Another drawback I see is that Travis starts running the next build as soon as workers are available, so the last workers in the old build might get data from the new build. I don't see this as a real issue though, as most of the time only the newest build in the same branch is relevant.

hpk42:
not sure i understand the secret management yet.  could you add a README somewhere that says how it's done?  and did you consider using "gitcrypt" for storing secrets?  i tentatively think it'd be good to standardize on it in various contexts.

fschulze:
Might make sense to use git-crypt, I haven't used it yet though.

For now the password is stored as an encrypted environment variable via https://travis-ci.org/devpi/devpi/settings and encrypted with gpg for each user. This also makes sure it doesn't show up in the travis logs.

PR #495, opened 2018-01-08
Properly set up HTTP basic auth during `devpi test`

vytas7:
* Use basic auth when downloading packages
 * Pass basic auth-enabled Pip index to tox

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -17,7 +17,7 @@ Pull Requests
 - If you are unsure on how to write a test, please ask and we will point you to
   existing tests which should help you writing new ones.
 - Embrace ``git rebase`` to create "nice" commits and history.
-- We strife for PEP8 compliance. New code should comply, old code must only
+- We strive for PEP8 compliance. New code should comply, old code must only
   comply when it's changed anyway. No unnecessary cleanups which would only
   make the PR harder to read.
 - Add a changelog entry in the ``news`` folder of touched packages as a new file.
--- a/client/devpi/test.py
+++ b/client/devpi/test.py
@@ -40,7 +40,9 @@ class DevIndex:
 
     def download_and_unpack(self, versioninfo, link):
         url = link.href
-        r = self.hub.http.get(url)
+        r = self.hub.http.get(url,
+                              auth=self.hub.current.get_basic_auth(url),
+                              cert=self.hub.current.get_client_cert(url))
         if r.status_code != 200:
             self.hub.fatal("could not receive", url)
         content = r.content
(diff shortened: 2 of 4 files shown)


vytas7:
The proposed patchset seems to solve the problem at hand, but I need advice on writing good unit tests for these, and Devpi tests are actually somewhat harder to understand than the code base itself :smiling_imp: 

Could something in spirit of ``client/testing/test_upload.py``: ``test_post_includes_auth_info`` do, just attaching fake basic auth and verifying the *requests* method has been called with *auth* / *cert* parameters?

fschulze:
@vytas7 Thanks! Yes, the ``test_post_includes_auth_info`` approach is totally fine.

I'm not sure if and where we should warn that the basic auth info will be in the uploaded test result data. If someone uses basic auth, then the test results are also behind that, so it shouldn't be an issue.

An alternative would be to create a ``pip.cfg`` in the temporary virtualenv which includes the basic auth info. I haven't tested that, but it might hide the basic auth info in that case.

vytas7:
Thanks for your input @fschulze !

I don't think it is a could idea to add unnecessary complexity by hacking in some temporary ``pip.cfg`` -- that would make it harder to manually rerun the commands from the report should the need arise.

In most common scenarios, it shouldn't be a problem users accessing reports with credentials since, as you point out, they would need either the same credentials, or another trusted way to authorize (such as internal network access like we do at work).
I added a warning for those rare scenarios where it is not the case, in a similar vein to what you get when configuring a client for a basic-auth protected Devpi.

If I were to gripe a bit about Devpi code base, I think testing could be made easier using [mock](https://docs.python.org/3/library/unittest.mock.html) (available on PyPi, native in Python 3.3+), [requests_mock](https://pypi.python.org/pypi/requests-mock) etc, but I did not want to make any substantial changes to the code base I am not very familiar with.

On the positive side, I have finally got the time to take a look at this again, and to finish the unit tests.

fschulze:
@vytas7 we use mock in some places already. Personally I try to avoid using it too much though, because I was bitten by it hiding changes in the mocked parts.
Some of the tools like requests_mock didn't exist or where not known when devpi tests were written.
The tests could be better in places, but as with most things they got complex with the growing/changing code base and there was no time/pressing need to refactor them yet.

vytas7:
> but as with most things they got complex with the growing/changing code base and there was no time/pressing need to refactor them yet.

Fully understood @fschulze , it is true for every large project with a living code base.

fschulze:
@vytas7 can you please add a changelog entry as described in https://github.com/devpi/devpi/blob/master/CONTRIBUTING.rst?

vytas7:
Sure, will fix!

I saw in the ``CHANGELOG`` that external contributors get thanked, shall I... umm thank myself? :smile:

fschulze:
@vytas7 you may do so :) We are still thinking about changing that, maybe adopt the way tox does it now.

vytas7:
@fschulze do not hesitate to let me know in case anything else if needed for this PR.

PR #497, opened 2018-01-09
Accept PyPy 5.10 version string

WolverineFan:
Fixes #496 (though it's kind of getting silly now).

PR #499, opened 2018-01-10
Fix issue496: PyPy 5.10 wheel upload failed

fschulze:
Fix issue496: PyPy 5.10 wheel upload failed because the version in the filename is longer again, the check for it is now removed, because it's pointless.

The extracted version string isn't correct (2.5.1.0 instead of 2.5.10) but it's not really used anyway besides as additional info in devpi-web.

This replaces PR #497 

cc @WolverineFan


gotcha:
lgtm

PR #506, opened 2018-01-19
487 url/history update in documentation

gotcha:


fschulze:
The functionality looks good. I tested it locally and it works.

I updated https://github.com/devpi/devpi/blob/master/CONTRIBUTING.rst#pull-requests to clarify how changelog fragment filenames should end, please update the filename.

The changelog for the anchor scrolling is missing.

The changelog entries must be in ``web/news`` to be properly picked up. Does this need to be reworded in ``CONTRIBUTING.rst`` or is it too easy to miss? Any improvement here is welcome.

gotcha:
Better this way ?

fschulze:
Looks good now!

PR #513, opened 2018-02-02
URLs and HTML according to PEP503

### LINKED ISSUE #494: Some Devpi pages not PEP 503 compliant
I know these are minor niggles, but in trying to describe to someone 'how to prove that a PyPI simple index or package URL is valid', I discovered that Devpi is, itself, not completely PEP 503 ( https://www.python.org/dev/peps/pep-0503/ ) compliant in terms of the HTML it returns for simple indexes. Among those I found:

* Simple Devpi indexes do not currently have a trailing ```</html>``` tag
* Simple indexes lead with ```<html>```, which is not valid HTML5 by the C3 validator (and by the PEP standard above); it should lead with ```<!DOCTYPE html>```
* Simple indexes should point to a package name with a trailing slash (```<a href="pip/">pip</a><br/>``` not ```<a href="pip">pip</a>```) to avoid unnecessary redirection that PEP 503 requires (see below)
* Redirection of the URL ```INDEX/+simple``` sends a user to ```INDEX/-simple```, a 404, which is presumably unintended behavior (I'd assume it's meant to redirect to ```INDEX/+simple/```?)

* Package listings *underneath* the simple index should lead with ```<!DOCTYPE html>```, not ```<html>``` as per validation above
* If directed to the URL ```.../+simple/package```, the user should be redirected to the URL ```.../+simple/package/``` as per PEP 503

Thanks for looking into these.

gotcha:
Fixes #494

### DIFF
--- a/server/devpi_server/main.py
+++ b/server/devpi_server/main.py
@@ -328,9 +328,12 @@ class XOM:
         pyramid_config.add_route("/+login", "/+login", accept="application/json")
         pyramid_config.add_route("/{user}/{index}/+e/{relpath:.*}", "/{user}/{index}/+e/{relpath:.*}")
         pyramid_config.add_route("/{user}/{index}/+f/{relpath:.*}", "/{user}/{index}/+f/{relpath:.*}")
+        pyramid_config.add_route("/{user}/{index}/+simple", "/{user}/{index}/+simple")
         pyramid_config.add_route("/{user}/{index}/+simple/", "/{user}/{index}/+simple/")
         pyramid_config.add_route("/{user}/{index}/+simple/{project}",
-                                 "/{user}/{index}/+simple/{project:[^/]+/?}")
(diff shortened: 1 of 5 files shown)


skylerbunny:
I did a pull of this branch on #513 and put it through quick paces. Determinations made:

All MUSTs on the PEP 503 page appear to be addressed.
All of the cases I mentioned (which amount to the MUSTs) appear to be addressed, against a local index. The SHOULD of redirection of a non-slash-ending URL to a slash-ending-URL is addressed.
I didn't specifically check mirrors (e.g., the default mirror), though I expect if they use the same code to generate +simple in their respective indexes, they should be fine.

There are of course other SHOULDs in 503, like checksums, but I think it's safe to say devpi's simple indexes are at least made 'compliant' by this PR, and additions like checksums could be construed as a future feature request.

Nice job.

PR #519, opened 2018-02-27
Add support for simple as well as +simple

TomTaylorLondon:
Artifactory and other tools enforce the 'simple' endpoint. This PR adds support for both '+simple' and 'simple' endpoints for basic functionality.

### DIFF
--- a/server/devpi_server/_index_refresh_form
+++ b/server/devpi_server/_index_refresh_form
@@ -393,6 +393,7 @@ class PyPIView:
     # index serving and upload
     #
 
+    @view_config(route_name="/{user}/{index}/simple/{project}")
     @view_config(route_name="/{user}/{index}/+simple/{project}")
     def simple_list_project_redirect(self):
         """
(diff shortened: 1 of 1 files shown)


blairham:
+1

fschulze:
Thanks for the PR, but there are several problems with it.

1. The commit message has nothing to do with the content
2. The referenced routes are missing
3. The referenced route name for ``simple_list_all`` is missing the trailing slash
4. There are no tests for the new routes
5. This would block ``simple`` as a package name, which would be a problem for someone who already has such a package (unlikely, but from past experience possible)

I'd say this is better suited as a plugin using the [``devpiserver_pyramid_configure`` hook](https://github.com/devpi/devpi/blob/master/server/devpi_server/hookspecs.py#L35). Which can be installed by anyone who needs this.

Another alternative is trying to fix Artifactory to allow any URLs, because enforcing a fixed one seems arbitrary in this case.

TomTaylorLondon:
Thanks for the quick response, I'll close this (agree with all points above - it was more for discussion) and I'll try the devpiserver_pyramid_configure route.

fschulze:
As a guide, this may be helpful. it's a minimal plugin adding two views:
https://github.com/devpi/devpi/tree/master/debugging

PR #531, opened 2018-04-17
 Update defaults and references to pypi.org

fschulze:


fschulze:
The postgresql test failures are caused by usage of old devpi-server releases and the resulting URL mismatch due to the pypi.org switch. They don't happen on my fork or once merged, because the current devpi-server version will be uploaded to the temporary devpi.net index.

PR #533, opened 2018-04-18
Remove vendored verlib.py

olasd:
Hi,

While packaging devpi for Debian, I'm reviewing the licensing situation of all files. I've stumbled upon the _verlib.py vendored file, whose license hasn't been set explicitly upstream.

As this library seemingly hasn't been used for years (the last usage I could find has been removed in f66477a31543de26de8e200becab32f5c5d6e809), I'm submitting this PR to clean up the useless module.

Thanks for considering!


fschulze:
Looks good, will merge when the tests passed.

Thanks for packaging!

Are you adding systemd configuration for process control? We would love to update our example config and properly document it. Currently it's using the deprecated ``--start`` option. It should use devpi in foreground mode instead.

fschulze:
The AppVeyour build fails because of this: https://github.com/jonparrott/cmarkgfm/issues/2

The postgresql issues are unrelated, I can reproduce them locally.

fschulze:
For the record, the postgresql errors happen because the fix required for pytest >= 3.4.0 isn't released yet and travis uses only released versions for testing PRs. The tests on master after merge are fine, because the pre release packages are generated and uploaded to devpi.

PR #544, opened 2018-05-20
Adjustable request timeout

evemorgen:
Hi!

This PR provides a way to adjust default request timeout, since waiting 30s x 4 times retry takes forever to timeout. New optional mirror arg has been added `--request-timeout` which defaults to 30 seconds when not specified, so this change shouldn't break anything.

I've updated news, docs and wrote some tests. Is there anything more to be done? 😄

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,4 +69,5 @@ ENV/
 /.hg/
 /etc/
 /tmp/
-
+*.sw[po]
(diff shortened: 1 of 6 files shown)


evemorgen:
Thanks!

fschulze:
You are welcome

PR #547, opened 2018-06-07
Fix exception when raising exception

awesomebytes:
`response.status` does not exist so it raises an exception while trying to raise a different exception (effectively hiding the real cause of the exception).


fschulze:
I'll leave this open for now to remind me to look at writing a test.

Issue #525, opened 2018-03-23
Adding removing items from lists


fschulze:
Currently the only way to change lists in index options with the REST API and thus with devpi-client is to send the full updated list. This is problematic, because it's easy to make mistakes and there can be race conditions if several clients do it at the same time.

I propose to support the following syntax for devpi-client:
```
devpi index acl_upload+=new_user
devpi index acl_upload-=old_user
```

On the server side this would result in this json data to be sent via the REST API using the http PATCH method:
```
{'acl_upload+': 'new_user'}
```
```
{'acl_upload-': 'old_user'}
```

Current devpi-client sends old unchanged data as well, which would change, but old devpi-client could be supported on the devpi-server side with some additional sanity checks.

RonnyPfannschmidt:
i propose using the patch method and a json change specification instead of magical json with extra text bits added

RonnyPfannschmidt:
for context see http://jsonpatch.com/

fschulze:
That would make the REST API nice, but we would have to reflect those operations in devpi-client then somehow, any suggestions? I guess something based on the path spec mentioned in there makes sense, as one can then manipulate arbitrary nested settings as well.

fschulze:
Released with devpi-server 4.8.0.

PR #550, opened 2018-06-18
Implement #525: the PATCH methods of user and index now accept JSON Patch

fschulze:
[JSON Patch](http://jsonpatch.com/).

The support is rudimentary for now, some methods like ``move`` are not supported.


fschulze:
Maybe we should add a field in the ``+api`` data to indicate the support for this feature. I think a ``features`` list of strings would be most useful, as that can also be used for plugins that add functionality that should be exposed to devpi-client plugins.

fschulze:
I'll rework this.

PR #554, opened 2018-06-27
530 pkg cache headers

fschulze:
This is unfortunately a blind shot for now. I copied the headers from pypi.org. I tried to verify this with pip, but the CacheControl code is initialized and then never triggered. I think I saw lots of caching related info when using ``-vvv`` during install with older pip versions, but with 9.0.3 and 10.0.1 I don't see that anymore. I have no idea what changed.

### DIFF
--- a/server/devpi_server/__init__.py
+++ b/server/devpi_server/__init__.py
@@ -1,1 +1,1 @@
-__version__ = '4.5.1.dev0'
+__version__ = '4.6.0.dev0'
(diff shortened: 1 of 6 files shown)


fschulze:
Ok, managed to verify this now. The issue was that I tried locally with http, but pip only caches when using https.

jaraco:
Thank you @fschulze and @gotcha. We verified this fix in our production environment as well. We're eager to see a release with this change!

PR #558, opened 2018-07-05
Implement #557: export directory symlinks

### LINKED ISSUE #557: Devpi does not upload symbolic links to directories
When building a namespace package with `python setup.py sdist`, it preserves symbolic links. But `devpi upload` creates a package that doesn't have them. That happens because `Checkout.export` only copies files, ignoring the symlinks even if they are in git.
Edit: it copies symlinks to files but not to the directories.

The project has Python and JS api that need access to the same data, so the data folder needs to be on the top level. The project has the structure of
```
/data/*.json
/namespace1/namespace2/{symlink to ../../data/}
/namespace1/namespace2/*.py
/js/*.js
```

lykahb:
This fixes #557. `python setup.py sdist` packages symlinks too.

### DIFF
--- a/client/devpi/upload.py
+++ b/client/devpi/upload.py
@@ -304,7 +304,7 @@ class Checkout:
         newrepo = basetemp.join(self.rootpath.basename)
         for fn in files:
             source = self.rootpath.join(fn)
-            if source.isfile():
+            if source.isfile() or source.islink():
                 dest = newrepo.join(fn)
                 dest.dirpath().ensure(dir=1)
                 source.copy(dest, mode=True)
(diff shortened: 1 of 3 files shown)


fschulze:
Can this generate packages where the link points outside the package? If so, I think we should at least warn people if it happens.

Besides that, an entry in client/news/557.bugfix would be nice. And I'd say it should be "Fix #557", since I'd consider it a bug.

lykahb:
Added the news file.

It's always been possible to have links outside of the package. The function isfile accepts the links to files as well. This PR only brings the ability to link the directories which is a minor change.

Checking the links may be helpful. Though, it has a larger scope and seems to belong rather to pip or setuptools. It'll affect the existing behavior too.

PR #559, opened 2018-07-09
Fix copying of symlinks

lykahb:
In the last PR I made a mistake assuming that `source.copy` would copy the symlink as-is. Instead it copied the directory. This PR contains the fix for that and improved tests. I'll add the news file if necessary or we can reuse the previous one.

Copying the repository manually is quite tricky. What do you think about using git shallow clone from directory instead?

### DIFF
--- a/client/devpi/upload.py
+++ b/client/devpi/upload.py
@@ -304,7 +304,11 @@ class Checkout:
         newrepo = basetemp.join(self.rootpath.basename)
         for fn in files:
             source = self.rootpath.join(fn)
-            if source.isfile() or source.islink():
+            if source.islink():
+                dest = newrepo.join(fn)
+                dest.dirpath().ensure(dir=1)
+                dest.mksymlinkto(source.readlink(), absolute=True)
+            elif source.isfile():
                 dest = newrepo.join(fn)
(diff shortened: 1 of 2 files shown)


fschulze:
Will merge once tests are through.

Regarding your proposal:
Changed but not checked in files need to be copied as well, not sure there are actually tests for that. Besides that, I'm all for an improved version. Maybe ``git archive`` is another alternative. I use it in another context to get a checked in version and pipe it directly into tar. Changed files would still have to be handled separately.

lykahb:
The tests fail for some unrelated reason
`TypeError: initproj() takes exactly 1 argument (2 given) `

fschulze:
Could you rebase with master? I fixed it there.

lykahb:
Thanks for the fix, the tests are good now.

lykahb:
@fschulze When is it possible to release the new version of devpi? The regular library release process in my company uses devpi from pypy. That's an obstacle for using that python-javascript library I described at #557 .

fschulze:
I've released devpi-client 4.0.3

lykahb:
Thank you!

PR #561, opened 2018-07-11
Add setting for max request body size (allow larger package upload)

tbartlett0:
Some people have a use-case of uploading very large (> 1GB) packages, see for example https://github.com/devpi/devpi/issues/342#issuecomment-290658626.

This patch adds an option to set max_request_body_size passed to waitress, allowing those users to set a higher maximum request size.

I note that in the linked issue there are comments suggesting that devpi loads the entire file contents into memory, so you may want to expand the help text to warn users. (I didn't observe any issues uploading a 2.5GB file, but with larger sizes or low-spec clients/servers there may be problems - I don't know.)

I have not created any test cases for this change as I'm unsure what tests would be of use -- I have manually tested that running with the option set to a low value causes a 413 error on upload; however this seems like something that would already be covered by waitress' tests...

### DIFF
--- a/server/devpi_server/config.py
+++ b/server/devpi_server/config.py
@@ -45,6 +45,10 @@ def addoptions(parser, pluginmanager):
             default=50,
             help="number of threads to start for serving clients.")
 
+    web.addoption("--max-request-body-size",  type=int,
+            default=1073741824,
+            help="maximum number of bytes in request body.")
+
     web.addoption("--outside-url",  type=str, dest="outside_url",
             metavar="URL",
             default=None,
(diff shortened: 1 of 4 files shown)


fschulze:
A useful test would be to check that the command line option value actually ends up being sent to waitress. Besides that I agree with you.

Could you add a changelog entry as described in https://github.com/devpi/devpi/blob/master/CONTRIBUTING.rst#pull-requests? Please let me know if you have trouble following the description, so we can improve it if necessary.

tbartlett0:
I've added a test to test_main, along the same lines as the test for the --threads option.

tbartlett0:
So the Travis build had an error in 'test_custom_data' for test_nginx_replica. It doesn't seem related to any of my changes.

PR #568, opened 2018-08-03
Mirror data fixes

fschulze:
This fixes the stored data in mirror indexes, so it can all be referenced and properly deleted.

### DIFF
--- a/server/devpi_server/extpypi.py
+++ b/server/devpi_server/extpypi.py
@@ -167,6 +167,42 @@ class PyPIStage(BaseStage):
             projects.add(project)
             self.key_projects.set(projects)
 
+    def del_project(self, project):
+        if not self.is_project_cached(project):
+            raise KeyError("project not found")
+        (is_fresh, links, cache_serial) = self._load_cache_links(project)
+        if links is None:
+            links = ()
+        entries = list(
+            x
+            for x in map(self._entry_from_cache_link, links)
+            if x.file_exists())
+        if not entries:
+            raise KeyError("no releases found")
+        for entry in entries:
+            entry.delete()
+        projects = self.key_projects.get(readonly=False)
+        if project in projects:
+            projects.remove(project)
+            self.key_projects.set(projects)
+
+    def del_entry(self, entry, cleanup=True):
+        project = entry.project
+        if project is None:
+            raise self.NotFound("no project set on entry %r" % entry)
+        if not entry.file_exists():
+            raise self.NotFound("entry has no file data %r" % entry)
+        entry.delete()
+        (is_fresh, links, cache_serial) = self._load_cache_links(project)
+        if links is None:
+            links = ()
+        links = ensure_deeply_readonly(list(filter(self._is_file_cached, links)))
+        if not links and cleanup:
+            projects = self.key_projects.get(readonly=False)
+            if project in projects:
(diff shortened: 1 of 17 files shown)


server/devpi_server/extpypi.py
```diff
@@ -157,6 +159,13 @@ def modify(self, index=None, **kw):
             self.ixconfig = newconfig
             return newconfig
 
+    def add_project(self, project):
```
gotcha: Rename function to ``update_project_names`` ?

server/devpi_server/extpypi.py
```diff
@@ -245,6 +254,9 @@ def _save_cache_links(self, project, links, serial):
         if old != data:
             threadlog.debug("saving changed simplelinks for %s: %s", project, data)
             key.set(data)
+            # add project if it isn't already, so deletion and offline mode
```
gotcha: Rephrase comment:

maintain list of currently cached project names to enable deletion and offline mode

server/devpi_server/model.py
```diff
@@ -679,6 +679,10 @@ def _set_versiondata(self, metadata):
         if version not in versions:
             versions.add(version)
             self.key_projversions(project).set(versions)
+        self.add_project(project)
+
+    def add_project(self, project):
```
gotcha: Same as above, rename function to ``update_project_names`` ?

server/devpi_server/extpypi.py
```diff
@@ -216,17 +217,18 @@ def _get_remote_projects(self):
 
     def list_projects_perstage(self):
         """ return set of all projects served through the mirror. """
+        if self.offline:
+            threadlog.warn("using stale projects list")
```
gotcha: Would it make sense to log ``offline mode: using stale projects list`` ?

server/devpi_server/extpypi.py
```diff
@@ -216,17 +217,18 @@ def _get_remote_projects(self):
...
         if self.cache_projectnames.is_fresh(self.cache_expiry):
             projects = self.cache_projectnames.get()
         else:
             # no fresh projects or None at all, let's go remote
             try:
                 projects = self._get_remote_projects()
             except self.UpstreamError:
-                if not self.cache_projectnames.exists():
-                    raise
                 threadlog.warn("using stale projects list")
```
gotcha: And express here what type of error happened...

server/devpi_server/extpypi.py
```diff
@@ -298,7 +300,9 @@ def get_simplelinks_perstage(self, project):
         """
         project = normalize_name(project)
         is_fresh, links, cache_serial = self._load_cache_links(project)
-        if is_fresh:
+        if is_fresh or self.offline:
```
gotcha: I do not understand test.offline being tested twice.
This might ne to be refactored to better express intent.

fschulze: This is changed in the 404 handling commit later on

fschulze:
I've made a final rebase to squash related commits.
I also added a new UpstreamNotFoundError exception class instead of using HTTPNotFound in the last commit about 404 handling.

fschulze:
I also did the is_fresh -> is_expired refactoring and added caching of 404s which makes this code much closer to before in behaviour.

PR #570, opened 2018-08-03
 Allow plugins to provide indexer backends for search.

fschulze:


web/devpi_web/whoosh_index.py
```diff
@@ -470,6 +477,71 @@ def query_projects(self, querystring, page=1):
...
+        for key, field in (('name', 'project'), ('summary', 'summary')):
+            value = fields.get(key, [])
+            if len(value) == 0:
+                continue
+            elif len(value) == 1:
+                parts.append('(type:%s "%s")' % (field, value[0].replace('"', '')))
+            else:
+                raise ValueError("Only one value allowed for query.")
+        querystring = (" %s " % operator).join(parts)
+        log.debug("xmlrpc_search {0}".format(querystring))
```
fschulze: Rename the log content, as this is the query string after transformation for the search backend.

PR #571, opened 2018-08-03
Fix #565: Don't list toxresults by default

### LINKED ISSUE #565: 'devpi list <package_name>' shows too detailed data
devpi-client version: 4.0.3

When invoking `devpi list <package_name>` devpi client actually lists few latest release files together with the test results. Fetching all tox results requires quite some time (especially when results are big). 

Would it make sense to display by default only the release files? User might want to view release files and tox results by adding -v flag: `devpi list <package_name> -v`.

fschulze:
@r2dan does this look good?

### DIFF
--- a/client/devpi/__init__.py
+++ b/client/devpi/__init__.py
@@ -1,2 +1,2 @@
 
-__version__ = '4.0.4.dev0'
+__version__ = '4.1.0.dev0'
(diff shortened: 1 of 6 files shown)


fschulze:
Not sure this deserves a major version bump. Strictly speaking it's a backward incompatible change, but the output isn't easily machine readable, so I doubt it's used in scripts for example.

r2dan:
Looks good. Regarding version bump - it's not an urgent thing so you can wait and release it with some other backward incompatible changes in the future.

gotcha:
Are there already other easily machine readable ways to access the same info ?
If yes, the risk of no version bump is minor. If not, I would wait for version bump.

fschulze:
The json info from the server is the best source for machine readable data.

PR #572, opened 2018-08-03
Remove scraping/crawling

### LINKED ISSUE #518: Confirm that crawling from PyPI is not needed anymore
Since external links are not enabled on PyPI, we might remove the crawling code: `perform_crawling` from `extpypi.py` might be removed.

fschulze:
Fix #518: There are no URLs on PyPI anymore that need to be scraped or crawled, so the code for that was removed.

### DIFF
--- a/server/devpi_server/filestore.py
+++ b/server/devpi_server/filestore.py
@@ -70,7 +70,6 @@ class FileStore:
                 basename=basename)
         entry = FileEntry(self.xom, key, readonly=False)
         entry.url = link.geturl_nofragment().url
-        entry.eggfragment = link.eggfragment
         # verify checksum if the entry is fresh, a file exists
         # and the link specifies a checksum.  It's a situation
         # that shouldn't happen unless some manual file system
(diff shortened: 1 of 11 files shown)


server/devpi_server/filestore.py
```diff
@@ -148,6 +144,7 @@ def __init__(self, msg, code=None, url=None):
 class FileEntry(object):
     BadGateway = BadGateway
     hash_spec = metaprop("hash_spec")  # e.g. "md5=120938012"
+    # BBB keep this for a while (1 year?) for older devpi-web
```
gotcha: I do not understand the comment

fschulze: When removing the eggfragment metaprop all versions of devpi-web released so far will break. I'm uncertain whether that is an issue or not, so I left it in for now.

web/devpi_web/views.py
```diff
@@ -239,7 +239,8 @@ def get_files_info(request, linkstore, show_toxresults=False):
     for link in sorted(filedata, key=attrgetter('basename')):
         url = url_for_entrypath(request, link.entrypath)
         entry = link.entry
-        if entry.eggfragment:
+        if getattr(entry, 'eggfragment', None):
+            # BBB for older devpi-server (<5.0.0)
```
gotcha: `# before 5.0.0, eggfragment value was the result of searching for downloads outside PyPI via scraping`

PR #573, opened 2018-08-04
Implement configuration file support

### LINKED ISSUE #502: Add option for configuration file
Currently all settings need to be passed as options on the command line. We should add an option to use a configuration file. I suggest to look into how pip handles that, as it allows all command line options to be supplied via a config file.

We should also look into whether we should support default locations or if it exclusively works via the command line option.

There currently is some mix of state and configuration kept in ``.nodeinfo`` which must also be handled differently.

This change is needed to remove state from the filesystem to fully implement #373

fschulze:
Fixes #502

### DIFF
--- a/client/testing/test_quickstart.py
+++ b/client/testing/test_quickstart.py
@@ -10,7 +10,7 @@ def test_dryrun(cmd_devpi):
 
 @pytest.mark.skipif("config.option.fast")
 def test_functional(cmd_devpi, monkeypatch, tmpdir):
-    monkeypatch.setenv("DEVPI_SERVERDIR", tmpdir.join("server"))
+    monkeypatch.setenv("DEVPISERVER_SERVERDIR", tmpdir.join("server"))
     monkeypatch.setenv("DEVPI_CLIENTDIR", tmpdir.join("client"))
     cmd_devpi("quickstart")
     try:
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -30,7 +30,7 @@ USER_MAN_CHAPTERS = commands\
           
                                  
 #export DEVPI_CLIENTDIR=$(CURDIR)/.tmp_devpi_user_man/client
-#export DEVPI_SERVERDIR=$(CURDIR)/.tmp_devpi_user_man/server
+#export DEVPISERVER_SERVERDIR=$(CURDIR)/.tmp_devpi_user_man/server
 
 chapter = commands
 
--- a/doc/quickstart-server.rst
+++ b/doc/quickstart-server.rst
@@ -38,6 +38,30 @@ And let's check the version::
 
 .. _genconfig:
 
+
+Using a configuration file for devpi-server
+-------------------------------------------
+
+A `strict YAML`_ conform configuration file can be used in place of the command line options.
+
+The configuration file will be looked for in your systems and your users default configuration location using the `appdirs`_ library.
+
+With the `-c/--configfile` option the location can be provided explicitly.
+
+The configuration file must start with a mapping and all devpi-server settings must be placed below the `devpi-server` key.
+For example:
+
+.. code-block:: yaml
+
+    devpi-server:
+      serverdir: /var/db/devpi-server
+      host: localhost
+      port: 8080
+
+.. _`strict YAML`: https://pypi.org/project/strictyaml/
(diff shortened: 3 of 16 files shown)


client/testing/test_quickstart.py
```diff
@@ -10,7 +10,7 @@ def test_dryrun(cmd_devpi):
 
 @pytest.mark.skipif("config.option.fast")
 def test_functional(cmd_devpi, monkeypatch, tmpdir):
-    monkeypatch.setenv("DEVPI_SERVERDIR", tmpdir.join("server"))
+    monkeypatch.setenv("DEVPISERVER_SERVERDIR", tmpdir.join("server"))
     monkeypatch.setenv("DEVPI_CLIENTDIR", tmpdir.join("client"))
```
gotcha: why is this not changed to ``DEVPICLIENT_CLIENTDIR``?

doc/Makefile
```diff
@@ -30,7 +30,7 @@ USER_MAN_CHAPTERS = commands\
           
                                  
 #export DEVPI_CLIENTDIR=$(CURDIR)/.tmp_devpi_user_man/client
```
gotcha: why is this not changed to ``DEVPICLIENT_CLIENTDIR`` ?

server/devpi_server/config.py
```diff
@@ -240,52 +242,143 @@ def try_argcomplete(parser):
...
+    config_dirs = config_dirs.split(os.pathsep)
+    config_dirs.append(
+        appdirs.user_config_dir('devpi-server', 'devpi'))
+    config_files = []
+    for config_dir in config_dirs:
+        config_file = os.path.join(config_dir, 'devpi-server.yml')
+        if os.path.exists(config_file):
+            config_files.append(config_file)
+    if len(config_files) > 1:
+        log.warn("Multiple configuration files found:\n%s", "\n".join(config_files))
```
gotcha: I would mention which file will be actually used.

server/devpi_server/config.py
```diff
@@ -240,52 +242,143 @@ def try_argcomplete(parser):
...
+            return {}
+        elif config.is_sequence():
+            raise InvalidConfigError("The config file must be a mapping, not a sequence.")
+        if 'devpi-server' not in config:
+            return {}
+        config = config['devpi-server']
+        if config.is_scalar():
+            return {}
+        elif config.is_sequence():
+            raise InvalidConfigError("The config file must be a mapping, not a sequence.")
```
gotcha: Same message as just before ?

server/devpi_server/config.py
```diff
@@ -240,52 +242,143 @@ def try_argcomplete(parser):
...
+    parser.post_process_actions(defaultget=defaultget)
+    args = parser.parse_args(argv[1:])
+    if args.help is True:
+        parser.print_help()
+        parser.exit()
     config = Config(args, pluginmanager=pluginmanager)
     return config
 
+
+def get_action_long_name(action):
```
gotcha: add a comment where this comes from

server/devpi_server/config.py
```diff
@@ -240,52 +242,143 @@ def try_argcomplete(parser):
...
-        return opt
-
-    def _processopt(self, opt):
-        try:
-            opt.default = self._defaultget(opt.dest)
-        except KeyError:
-            pass
-        if opt.help and opt.default:
-            opt.help += " [%s]" % opt.default
+    def post_process_actions(self, defaultget=None):
```
gotcha: add a comment about what it does

doc/quickstart-server.rst
```diff
@@ -38,6 +38,30 @@ And let's check the version::
...
+
+    devpi-server:
+      serverdir: /var/db/devpi-server
+      host: localhost
+      port: 8080
+
+.. _`strict YAML`: https://pypi.org/project/strictyaml/
+.. _`appdirs`: https://pypi.org/project/appdirs/
+
+
```
gotcha: no mention of environment variables

client/testing/test_quickstart.py
fschulze: This PR only adds configuration file support to devpi-server. If we ever add it to devpi-client, then it should be updated there.

PR #574, opened 2018-08-04
Description is rendered on demand.

fschulze:


fschulze:
Doh, I wondered where the test failure came from. It's because for PRs the final releases from PyPI are used, not the current development versions via devpi.net.

Issue #504, opened 2018-01-19
Doczip unpacking should work like a cache


fschulze:
Currently the doczip unpacking is solely handled via events. It should work more like a cache, so the unpacking happens immediately on request if it doesn't exist. This would also allow the unpacking be redone if it's missing on the filesystem.

This change is needed to remove persistent state from the filesystem to fully implement #373

fschulze:
The latter

PR #576, opened 2018-08-06
Implement #504: the documentation is unpacked and cached on demand

fschulze:


web/devpi_web/whoosh_index.py
```diff
@@ -281,7 +281,7 @@ def _update_projects(self, writer, projects, clear=False):
                 if '+doczip' not in project:
                     continue
                 if not project['+doczip'].exists():
-                    log.debug("documentation not unpacked for %s", data['path'])
+                    log.error("documentation zip doesn't exist for %s", data['path'])
```
gotcha: Rather "documentation zip file is missing" ?

PR #580, opened 2018-08-30
Allow deletion of specific package versions on a mirror.

fschulze:
Only the downloaded files will be removed.
The file will be fetched again from the mirror source on request.

### DIFF
--- a/server/devpi_server/extpypi.py
+++ b/server/devpi_server/extpypi.py
@@ -181,6 +181,21 @@ class PyPIStage(BaseStage):
             self.cache_retrieve_times.expire(project)
             self.key_projects.set(projects)
 
+    def del_versiondata(self, project, version, cleanup=True):
+        project = normalize_name(project)
+        if not self.has_project_perstage(project):
+            raise self.NotFound("project %r not found on stage %r" %
+                                (project, self.name))
+        # since this is a mirror, we only have the simple links and no
+        # metadata, so only delete the files and keep the simple links
+        # for the possibility to re-download a release
+        (is_expired, links, cache_serial) = self._load_cache_links(project)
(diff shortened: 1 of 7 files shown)


server/devpi_server/extpypi.py
```diff
@@ -181,6 +181,19 @@ def del_project(self, project):
             self.cache_retrieve_times.expire(project)
             self.key_projects.set(projects)
 
+    def del_versiondata(self, project, version, cleanup=True):
+        project = normalize_name(project)
+        if not self.has_project_perstage(project):
+            raise self.NotFound("project %r not found on stage %r" %
+                                (project, self.name))
+        # since this is a mirror, we only delete the actual files
```
gotcha: Better comment:
As this is a mirror, there is no metadata. Hence, only delete mirrored files.

server/test_devpi_server/test_views.py
```diff
@@ -1694,6 +1691,91 @@ def test_delete_from_mirror(mapp, pypistage, testapp):
         assert not other_entry.file_exists()
 
 
+def test_delete_version_from_mirror(mapp, pypistage, testapp):
+    mapp.login_root()
```
gotcha: add comment with high level use case

server/test_devpi_server/test_views.py
```diff
@@ -1694,6 +1691,91 @@ def test_delete_from_mirror(mapp, pypistage, testapp):
...
+    mapp.use("root/pypi")
+    name = "pytest"
+    other_mirrorpath = "/%s-2.5.zip" % name
```
gotcha: rename variables to path25 path26 link25 link26 entry25 entry26

PR #588, opened 2018-09-18
Fix empty description

fschulze:
Fix rendering of empty description which caused a traceback on Python 2.7.

### DIFF
--- a/web/devpi_web/description.py
+++ b/web/devpi_web/description.py
@@ -1,16 +1,8 @@
 from __future__ import unicode_literals
-from devpi_common.validation import normalize_name
 import io
 import py
 import readme_renderer.rst
 import readme_renderer.txt
-
-
-def get_description_file(stage, name, version):
-    config = stage.xom.config
-    name = normalize_name(name)
-    return config.serverdir.join(
-        '.web', stage.user.name, stage.index, name, version, 'description.html')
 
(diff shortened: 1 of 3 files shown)


web/devpi_web/description.py
```diff
@@ -28,11 +20,13 @@ def get_description(stage, name, version):
...
-        html = '<p>No description in metadata</p>'
-    else:
+    if desc:
         html = render_description(stage, desc)
-    return py.builtin._totext(html, "utf-8")
+    else:
+        html = '<p>No description in metadata</p>'
+    if not py.builtin._istext(html):
+        html = py.builtin._totext(html, "utf-8")
+    return html
```
hpk42: i btw think render_description should return unicode, and here we could not do the check (or just an assertion that it is unciode).

fschulze: I dislike asserts in production code, as they can be disabled with -o. With the check we don't care what we get back, we will always return unicode. I was just under the assumption that _totext would handle that already, but the _istext check is used in other parts of devpi as well.

fschulze:
The failures for postgresql are because the devpi-web tests are run against the previous release. It's fine when the current releases are used, see 
https://travis-ci.org/fschulze/devpi/builds/429927630

PR #590, opened 2018-09-28
Use server side operations for modifying indexes

fschulze:
This prevents undoing changes if two clients modify the same index concurrently.
Additionally it's easier to add and remove items from list values by using ``+=`` and ``-=``.

### DIFF
--- a/client/devpi/index.py
+++ b/client/devpi/index.py
@@ -76,12 +76,6 @@ def main(hub, args):
 
 def parse_keyvalue_spec_index(hub, keyvalues):
     try:
-        kvdict = parse_keyvalue_spec(keyvalues)
+        return parse_keyvalue_spec(keyvalues)
     except ValueError:
         hub.fatal("arguments must be format NAME=VALUE: %r" %( keyvalues,))
-    # XXX devpi-server 3.0.0 handles the splitting on it's own, this is for
-    # compatibility with older devpi-server
-    for key in ("acl_upload", "bases", "pypi_whitelist"):
-        if key in kvdict:
-            kvdict[key] = [x for x in kvdict[key].split(",") if x]
-    return kvdict
--- a/client/devpi/index.py
+++ b/client/devpi/index.py
@@ -1,16 +1,30 @@
-from devpi.use import parse_keyvalue_spec
+from devpi.use import get_keyvalues
+
 
 def index_create(hub, url, kvdict):
     hub.http_api("put", url, kvdict)
     index_show(hub, url)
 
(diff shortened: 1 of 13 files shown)


client/devpi/index.py
```diff
@@ -48,13 +62,14 @@ def parse_posargs(hub, args):
     indexname = args.indexname
     keyvalues = list(args.keyvalues)
     if indexname and "=" in indexname:
-        keyvalues.append(indexname)
```
fschulze: Reminder to self, add a comment on why we switch from append to insert

client/news/525.feature
```diff
@@ -0,0 +1 @@
+Use server side operations for modifying indexes if the devpi-server version supports it. This prevents undoing changes if two clients modify the same index concurrently.
```
fschulze: Change "undoing" to something like "last one wins and overwrites previous change"

client/testing/test_functional.py
```diff
@@ -281,29 +282,30 @@ def test_switch_preserves_auth(out_devpi, url_of_liveserver, url_of_liveserver2)
...
+    yield user_id
+    try:
+        mapp.delete_user(user=user_id, code=201)
+    except:
+        pass
+
+
+@pytest.fixture
+def existing_user_id(new_user_id, mapp):
+    """ Create a temporary user - will be deleted when the
```
fschulze: Fix docstring, it mentions parts that are actually done in new_user_id

client/testing/test_use.py
```diff
@@ -506,16 +506,42 @@ def test_main_setcfg(self, scheme, basic_auth, mock_http_api, cmd_devpi, tmpdir,
...
+@pytest.mark.parametrize("input expected".split(), [
+    (["hello=123", "world=42"], [("default", "hello", "123"), ("default", "world", "42")]),
+    (["hello=123=1"], [("default", "hello", "123=1")]),
+    (["hello=1", "hello=2"], [("default", "hello", "1"), ("default", "hello", "2")]),
+    (["hello:=1"], [("replace", "hello", "1")]),
+    (["hello+=1"], [("add", "hello", "1")]),
+    (["hello-=1"], [("remove", "hello", "1")]),
+    (["hello?=1"], [("test", "hello", "1")]),
+    (["hello+=1", "hello-=2"], [("add", "hello", "1"), ("remove", "hello", "2")])])
+def getparse_keyvalues_actions(input, expected):
```
fschulze: Rename to ``test_getparse_keyvalues_actions`` and fix the parameters.

PR #593, opened 2018-10-03
WIP: Mirror search original case names

### LINKED ISSUE #592: `pip search` using devpi is case sensitive, while according to PEP 426 it should not be
I am not sure if I titled this issue correctly, but I definitely see an issue in devpi's search behaviour.

Compare results of `pip search uwsgi` on a virtualenv NOT using devpi:
```
(venv_test_search) gdubicki@mac ~ $ pip search uwsgi
uWSGI (2.0.17.1)                   - The uWSGI server
  INSTALLED: 2.0.17.1 (latest)
uwsgi-cloudwatch (0.0.1.2)         - uwsgi-cloudwatch
(...)
```
...with the one using it:
```
(venv) [root@versions.XXX:/srv/eg-versions/current]# pip search uwsgi
uwsgi (2.0.17.1)              - [root/pypi]
uwsgi-env ()                  - [root/pypi]
(...)
```

Similarly to #455 this breaks behaviour of python puppet module because it expects LATEST/INSTALLED string to be in the output.

And I also think that this is invalid according to PEP 426.

gdubicki:
Fixed #592

Note: this is work in progress, please do not review yet. I just want to get a first opinion about direction I am going.

### DIFF
--- a/common/devpi_common/validation.py
+++ b/common/devpi_common/validation.py
@@ -15,7 +15,7 @@ def normalize_name(name):
     Any runs of non-alphanumeric/. characters are replaced with a single '-'.
     """
     name = ensure_unicode(name)
-    return safe_name_rex.sub('-', name).lower()
+    return safe_name_rex.sub('-', name)
 
 def safe_version(version):
     """Convert an arbitrary string to a standard version string
--- a/common/devpi_common/vendor/_pip.py
+++ b/common/devpi_common/vendor/_pip.py
@@ -55,8 +55,15 @@ class HTMLPage(object):
         class AnchorParser(html_parser.HTMLParser, object):
             def __init__(self, *args, **kwargs):
                 super(AnchorParser, self).__init__(*args, **kwargs)
+                # dict of pairs: (a_tag_data, a_tag_attributes)
                 self.anchors = {}
                 self.last_href = None
+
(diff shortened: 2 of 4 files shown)


common/devpi_common/validation.py
```diff
@@ -15,7 +15,7 @@ def normalize_name(name):
     Any runs of non-alphanumeric/. characters are replaced with a single '-'.
     """
     name = ensure_unicode(name)
```
fschulze: This function may not be changed, it's essential for all devpi internals.

fschulze:
Thanks for looking into this! Unfortunately the current approach is on the wrong track, see my review comments.

gdubicki:
Thanks for the hints @fschulze but after spending some time on this I feel that it's too hard for a person not knowing that code well - like myself - to do it in a reasonable amount of time. Can you try to finish it?

gdubicki:
I have pushed some comments to make the link parsing part easier for you to use.

fschulze:
I have no ETA for fixing this. The companies with service contracts have other priorities and I have no free time for this ATM. If anyone would like to sponsor this, contact us through https://merlinux.eu

fschulze:
I have investigated this further and the necessary changes go quite deep and will even require db schema changes.

fschulze:
As this requires schema changes it needs to be visited again for the next major release and the required changes will need a fresh start.

PR #596, opened 2018-10-04
Handle mirrors that don't return a correct X-PYPI-LAST-SERIAL header

elesel:
I'm using a PyPI mirror provides a "None" value for the X-PYPI-LAST-SERIAL in its responses, and I thought others might run into the same issue.

### DIFF
--- a/server/devpi_server/extpypi.py
+++ b/server/devpi_server/extpypi.py
@@ -416,7 +416,10 @@ class PyPIStage(BaseStage):
         # Returning a 200 with "no such project" was originally meant to
         # provide earlier versions of easy_install/pip to request the full
         # simple page.
-        serial = int(response.headers.get(str("X-PYPI-LAST-SERIAL"), "-1"))
+        try:
+            serial = int(response.headers.get(str("X-PYPI-LAST-SERIAL")))
+        except:
+            serial = -1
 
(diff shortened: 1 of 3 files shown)


server/devpi_server/extpypi.py
```diff
@@ -416,7 +416,10 @@ def get_simplelinks_perstage(self, project):
         # Returning a 200 with "no such project" was originally meant to
         # provide earlier versions of easy_install/pip to request the full
         # simple page.
-        serial = int(response.headers.get(str("X-PYPI-LAST-SERIAL"), "-1"))
+        try:
+            serial = int(response.headers.get(str("X-PYPI-LAST-SERIAL"), "-1"))
```
fschulze: I'd remove the "-1". That way the code path is the same way for any missing or invalid header combination.

fschulze:
Thanks for this! For the changelog entry I'd classify this as a bugfix.

elesel:
I believe I've done what you asked. Please let me know if you need anything else.

fschulze:
I've expanded a test and made some minor changes. Will merge after test run.

fschulze:
The test failures come from Pyramid 1.10a1 and will be fixed with the next release.

elesel:
Thanks!

Issue #598, opened 2018-10-08
Bug: Seriously inefficient byte concatenation hindering and breaking streamed downloads


DomHudson:
## Summary

I am experiencing issues pulling large packages from my root/pypi mirror. The downloads start fast but progressively get slower over time. Eventually the download will become very very slow and never complete. The issue is most prevalent to me with the Tensorflow package where the wheel is large - approximately 60MB.

## The cause

I have investigated this issue and believe it is caused by inefficient byte concatenation in [`filestore.iter_cache_remote_file`](https://github.com/devpi/devpi/blob/master/server/devpi_server/filestore.py#L298)

Repeatedly concatenating these byte-strings is triggering excessive garbage collection or memory reallocation.

The following is a working example of this, this code is extracted from `devpi_common.request` and `devpi_server.filestore`. 

The original code gets progressively slower as it loops to around iteration 3971 before becoming unresponsive. This takes approximately 40 seconds on my machine.

The fixed code runs at a linear speed and completes the entire download (6148 iterations) in 9 seconds.

## Original code
This code is almost identical to what is currently in devpi. I just extracted it to prove the problem.
```python
import sys
from requests import Session
from requests.adapters import HTTPAdapter
from requests.exceptions import ConnectionError, RequestException, BaseHTTPError, SSLError

class RetrySession(Session):
    def __init__(self, max_retries):
        super(RetrySession, self).__init__()
        if max_retries is not None:
            self.mount('https://', HTTPAdapter(max_retries=max_retries))
            self.mount('http://', HTTPAdapter(max_retries=max_retries))

def new_requests_session(agent=None, max_retries=None):
    if agent is None:
        agent = "devpi"
    else:
        agent = "devpi-%s/%s" % agent
    agent += " (py%s; %s)" % (sys.version.split()[0], sys.platform)
    session = RetrySession(max_retries)
    session.headers["user-agent"] = agent
    session.ConnectionError = ConnectionError
    session.RequestException = RequestException
    session.Errors = (RequestException, BaseHTTPError)
    session.SSLError = SSLError
    return session

session = new_requests_session()

URL = 'https://files.pythonhosted.org/packages/ce/d5/38cd4543401708e64c9ee6afa664b936860f4630dd93a49ab863f9998cd2/tensorflow-1.11.0-cp36-cp36m-manylinux1_x86_64.whl'

r = session.get(
    URL,
    stream = True,
    allow_redirects = True,
    timeout = 5
)

content = b''
i = 0
[...]

PR #599, opened 2018-10-08
Bug fix: Seriously inefficient byte concatenation hindering and breaking streamed downloads #598

DomHudson:
## Summary
This is a bug fix for my issue [Bug: Seriously inefficient byte concatenation hindering and breaking streamed downloads #598](https://github.com/devpi/devpi/issues/598)

This should significantly speed up all downloads from pypi (providing they are over the `10240` size) but expontentially affects larger packages.  It also reduces memory and CPU usage.

Please see the issue for more information, but it can be summarised as so:

The original code gets progressively slower in each iteration of this loop (causing large packages to eventually fail).
The fixed code runs at a linear speed and completes the entire download significantly faster.

### DIFF
--- a/server/devpi_server/filestore.py
+++ b/server/devpi_server/filestore.py
@@ -4,6 +4,7 @@ for all indexes.
 
 """
 from __future__ import unicode_literals
+from io import BytesIO
 import hashlib
 import mimetypes
 from wsgiref.handlers import format_date_time
@@ -294,15 +295,15 @@ class FileEntry(object):
 
         yield self._headers_from_response(r)
 
-        contents = []
+        content = BytesIO()
         while 1:
             data = r.raw.read(10240)
             if not data:
                 break
-            contents.append(data)
+            content.write(data)
             yield data
 
-        content = b''.join(contents)
+        content = content.getvalue()
 
         filesize = len(content)
         if content_size and int(content_size) != filesize:
@@ -365,15 +366,15 @@ class FileEntry(object):
 
         yield self._headers_from_response(r)
 
-        contents = []
+        content = BytesIO()
         while 1:
             data = r.raw.read(10240)
             if not data:
                 break
-            contents.append(data)
+            content.write(data)
             yield data
 
-        content = b''.join(contents)
+        content = content.getvalue()
 
         err = self.check_checksum(content)
         if err:
--- a/server/devpi_server/filestore.py
+++ b/server/devpi_server/filestore.py
@@ -294,13 +294,15 @@ class FileEntry(object):
 
         yield self._headers_from_response(r)
 
-        content = b''
+        contents = []
         while 1:
             data = r.raw.read(10240)
             if not data:
                 break
-            content = content + data
+            contents.append(data)
             yield data
+
+        content = b''.join(contents)
 
         filesize = len(content)
         if content_size and int(content_size) != filesize:
@@ -363,13 +365,15 @@ class FileEntry(object):
 
         yield self._headers_from_response(r)
 
-        content = b''
+        contents = []
         while 1:
             data = r.raw.read(10240)
             if not data:
                 break
-            content = content + data
+            contents.append(data)
             yield data
+
+        content = b''.join(contents)
 
         err = self.check_checksum(content)
         if err:
--- a/server/news/598.bugfix
+++ b/server/news/598.bugfix
@@ -1,0 +1,3 @@
+fix #598: streaming download now uses BytesIO to avoid performance issues for downloads with more than a few MB.
+
+Thanks to Dom Hudson from http://www.thoughtriver.com for the report and initial benchmark code.


fschulze:
cc @r2dan

r2dan:
@fshulze: just wondering if it fixes the problem we had with replication of the large files we discussed some time ago. From what I see symptoms are pretty much the same...

fschulze:
@r2dan that's why I cced you. If you have the chance, could your try the changes? The next release will probably be at the end of this month.

r2dan:
I'll try to reproduce this issue on my test infrastructure - I cannot promise that it will reproduce though.

DomHudson:
Thanks both. @r2dan you could try the code I posted in the issue - this replicates it for me without the overhead of a full installation.

Thanks,
Dom

r2dan:
@DomHudson : your example works perfectly fine. Regarding string concatenation efficiency, there are even more efficient methods according to this reading: https://waymoot.org/home/python_string/ (just FYI - speed improvement is not that big to invest time in the implementation). Thanks!

DomHudson:
On my environments it's a fatal bug - it is impossible to install tensorflow . It's a speed improvement that makes the difference between it functioning and not. Do you mean the unpatched code works fine? 

If so much be highly dependent on environment but the effect was very severe on my platforms.

r2dan:
I meant the fix works fine - usage of lists and join method.

DomHudson:
Sorry my mistake! I thought you were saying  this shouldn't be merged

fschulze:
I did some more tests.

I ran a SimpleHTTPServer locally with ``python -m SimpleHTTPServer 8000`` and files of various sizes with the following code:
```
from hashlib import sha256
from io import BytesIO
from requests import Session
from requests.adapters import HTTPAdapter
from requests.exceptions import ConnectionError, RequestException, BaseHTTPError, SSLError
from timeit import timeit
import sys


class RetrySession(Session):
    def __init__(self, max_retries):
        super(RetrySession, self).__init__()
        if max_retries is not None:
            self.mount('https://', HTTPAdapter(max_retries=max_retries))
            self.mount('http://', HTTPAdapter(max_retries=max_retries))


def new_requests_session(agent=None, max_retries=None):
    if agent is None:
        agent = "devpi"
    else:
        agent = "devpi-%s/%s" % agent
    agent += " (py%s; %s)" % (sys.version.split()[0], sys.platform)
    session = RetrySession(max_retries)
    session.headers["user-agent"] = agent
    session.ConnectionError = ConnectionError
    session.RequestException = RequestException
    session.Errors = (RequestException, BaseHTTPError)
    session.SSLError = SSLError
    return session


session = new_requests_session()

URL = 'http://localhost:8000/test6mb'


def method1():
    r = session.get(
        URL,
        stream=True,
        allow_redirects=True,
        timeout=5)
    content = b''
    while 1:
        data = r.raw.read(10240)
        if not data:
            break
        content = content + data
        yield data
    print(sha256(content).hexdigest())


def method2():
    r = session.get(
        URL,
        stream=True,
        allow_redirects=True,
        timeout=5)
    content = []
    while 1:
        data = r.raw.read(10240)
        if not data:
            break
        content.append(data)
        yield data
    content = b''.join(content)
    print(sha256(content).hexdigest())


def method3():
    r = session.get(
        URL,
        stream=True,
        allow_redirects=True,
        timeout=5)
    content = BytesIO()
    while 1:
        data = r.raw.read(10240)
        if not data:
            break
        content.write(data)
        yield data
    content = content.getvalue()
    print(sha256(content).hexdigest())


def run_method(method):
    for data in method():
        pass


print(timeit('run_method(method1)', number=10, globals=globals()))
print(timeit('run_method(method2)', number=10, globals=globals()))
print(timeit('run_method(method3)', number=10, globals=globals()))
```

I aborted ``method1`` testing above 10mb, because it was clearly bad.

For me ``method3`` was a tiny bit slower for packages smaller than 100mb, but from there ``method2`` became slower while ``method3`` scaled linearly up to 600mb where I stopped.

fschulze:
@DomHudson are you ok with my changes and the changelog entry? If so, I will merge after the tests are through.

DomHudson:
Hmm I'm having this issue again after updating from my fork/branch to the released version. I ran the same tests as you above and basically had the same conclusion. I have no idea why I seem to be getting this problem again but will do some further investigation.


image

fschulze:
Out of curiosity, how did you generate the graph? I never got to look into stuff like that till now, but always wanted to.

DomHudson:
Sure this is a [seaborn lineplot](https://seaborn.pydata.org/examples/errorband_lineplots.html) . Seaborn is like a wrapper over the top of matplotlib and makes it a lot easier to do common types of graphs

fschulze:
Thanks! By just a brief skimming of the tutorial I already learned a lot of stuff.

PR #600, opened 2018-10-13
Feature: Markdown description support

DomHudson:
## Summary
Adding support for using markdown in package descriptions.

Enabled by existing functionality in [`readme_renderer`](https://github.com/pypa/readme_renderer/pull/55)

## Use
Simply add markdown to the existing the `long_description` field and set the `long_description_content_type` field to 'text/markdown'.

## Closes
https://github.com/devpi/devpi/issues/193

## Notes
The tox tests are failing on my local with py2.7 but passing if I run them manually with pytest and the same version of python. I believe this is because it is installing the unmodified devpi-server as part of the spinning up the 2.7 which doesn't contain the additional `long_description_content_type` field but may be wrong - feedback welcome.

I was a little confused by the stuff going on with `py.bultin._istext` and `py.bultin._totext` [here](https://github.com/devpi/devpi/blob/master/web/devpi_web/description.py#L27). In some cases, it seemed to encode the html as utf-8 and then decode it straight back again.

I've left it in as `devpi_web.description.DescriptionRenderer._force_text`.

### DIFF
--- a/server/devpi_server/model.py
+++ b/server/devpi_server/model.py
@@ -621,7 +621,9 @@ class PrivateStage(BaseStage):
         'provides', 'requires', 'obsoletes',
         # Metadata 1.2
         'project_urls', 'provides_dist', 'obsoletes_dist',
-        'requires_dist', 'requires_external', 'requires_python')
+        'requires_dist', 'requires_external', 'requires_python',
+        # Metadata 2.1
+        'description_content_type', 'provides_extras')
     metadata_list_fields = (
         'platform', 'classifiers', 'obsoletes',
         'requires', 'provides', 'obsoletes_dist',
--- a/server/news/193.feature
+++ b/server/news/193.feature
@@ -1,0 +1,3 @@
+feature #193: ensuring that the `description_content_type` and `provides_extras` fields are handled.
+
+This is used to add support for using alternative content types in the package descriptions such as `text/markdown`.
--- a/web/devpi_web/description.py
+++ b/web/devpi_web/description.py
@@ -1,42 +1,108 @@
 from __future__ import unicode_literals
 import io
 import py
+import readme_renderer.markdown
 import readme_renderer.rst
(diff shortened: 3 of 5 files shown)


server/devpi_server/model.py
```diff
@@ -616,7 +616,7 @@ class PrivateStage(BaseStage):
         'metadata_version', 'summary', 'home_page', 'author', 'author_email',
         'maintainer', 'maintainer_email', 'license', 'description',
         'keywords', 'platform', 'classifiers', 'download_url',
-        'supported_platform', 'comment',
+        'description_content_type', 'supported_platform', 'comment',
```
fschulze: ``description_content_type`` is Metadata 2.1 according to https://github.com/pypa/twine/blob/0191a22a1ea51e4ff18fd20706094e03bb1abee4/twine/package.py#L145

I would add that and ``provides_extras`` as a separate commit.

I'll review the rest later.

fschulze: Also see 
https://www.python.org/dev/peps/pep-0566/

DomHudson:
Okay thanks. I've added change logs and separated the metadata 2.1 (with `provides_extras`) via it's own section in `metadata_list_fields`.

Thanks
Dom

fschulze:
I've rebase with master for an unrelated test failure fix and while at it squashed the commits into related ones.

DomHudson:
Great thanks - let me know if you have other changes or if this good to go

Thanks,
Dom

fschulze:
I removed the fixed time stuff in the tests, as they aren't needed apparently, maybe leftovers from development?

I also added a ``skipif`` marker to skip the test with older devpi versions, as we test with 4.2.1.

fschulze:
Now waiting for the test results at https://travis-ci.org/fschulze/devpi/builds/443769088, as PRs aren't run with pre-release versions of the other devpi packages.

PR #609, opened 2018-11-16
Miscellaneous: Ensuring pep8 newline standards

DomHudson:
## Summary
I automated this change to ensure pep-8 compliance with newlines:
1. Remove trailing whitespace
2. Remove whitespace on lines that contain no code
3. Ensure files end with a single trailing new line


fschulze:
Thanks for the effort. There are less changes than I thought. But we need a way to enforce them, so they don't creep in again. We would need to enable the pep8 checker, but add a list of current issues to ignore. We would also need a way to check the ``.py`` files in the ``doc`` folder.

In general we have the unwritten policy to only fix pep8 issues on lines touched in a PR to avoid too many changes at once and possible conflicts caused by that in branches still being worked on.

PR #612, opened 2018-12-03
make setup.py much more readable

GreatBahram:
Hi 
I just cleaned and made `setup.py` more readable

### DIFF
--- a/server/devpi_server/__version__.py
+++ b/server/devpi_server/__version__.py
@@ -1,0 +1,8 @@
+
+__title__ = 'devpi-server'
+__description__ = 'devpi-server: reliable private and pypi.org caching server'
+__url__ = 'http://doc.devpi.net'
+__version__ = '4.8.1.dev0'
+__maintainer__ = 'Kenneth Reitz'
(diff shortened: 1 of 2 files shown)


server/devpi_server/__version__.py
```diff
@@ -0,0 +1,8 @@
+
+__title__ = 'devpi-server'
+__description__ = 'devpi-server: reliable private and pypi.org caching server'
+__url__ = 'http://doc.devpi.net'
+__version__ = '4.8.1.dev0'
+__maintainer__ = 'Kenneth Reitz'
```
RonnyPfannschmidt: wtf

fschulze:
Thanks for the effort, but there are too many unrelated changes and the readability of setup.py isn't important. You also made changes which break installation.

Using ``exec`` during setup also isn't good practice.

When the time comes that we drop support for Python 2.7 we will switch to use a config file instead of setup.py. For now we will keep it as is.

server/devpi_server/__version__.py
GreatBahram: for sake of the time, I copied requests module and just change the values

PR #626, opened 2019-02-05
Added option '--root-passwd' which takes a string argument to be the …

apalsson:
This patch adds a new command line option, `--root-passwd`, which takes one string argument to use as the initial root password instead of an empty string.  
If the root account already exist, this option has no effect.

This change makes it easier to initially deploy the server in environments without interactive terminals, such as Docker where the initial password can be supplied via environment variables.

### DIFF
--- a/client/devpi/main.py
+++ b/client/devpi/main.py
@@ -733,7 +733,7 @@ def user(parser):
         help="user name")
     parser.add_argument("keyvalues", nargs="*", type=str,
         help="key=value configuration item.  Possible keys: "
-             "email, password.")
+             "email, password and pwhash.")
 
 
 @subcommand("devpi.user:passwd")
--- a/client/devpi/user.py
+++ b/client/devpi/user.py
@@ -18,7 +18,7 @@ def getnewpass(hub, username):
         hub.error("passwords did not match")
 
 def user_create(hub, user, kvdict):
-    if "password" not in kvdict:
+    if "password" not in kvdict and "pwhash" not in kvdict:
         kvdict["password"] = getnewpass(hub, user)
     hub.http_api("put", hub.current.get_user_url(user), kvdict)
     hub.info("user created: %s" % user)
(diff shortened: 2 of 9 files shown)


fschulze:
In my opinion this should only accept an already hashed password and not a clear text one, otherwise it's way too easy to have a leak in logs etc.
Would you be interested in extending the code accordingly?
I think ``create_user`` should detect a password hash, that way it's also possible to use it for creating other users. If possible (I haven't checked) the result sent to devpi-client should indicate that the hash and not the password itself was set.

apalsson:
> In my opinion this should only accept an already hashed password and not a clear text one, otherwise it's way too easy to have a leak in logs etc.
> Would you be interested in extending the code accordingly?
> I think `create_user` should detect a password hash, that way it's also possible to use it for creating other users. If possible (I haven't checked) the result sent to devpi-client should indicate that the hash and not the password itself was set.

I think you are right, however I rather the code be explicit than rely on detection. Therefore I added another option for setting an already hashed password.

fschulze:
Thanks! I hope to review and try this soon.

server/devpi_server/auth.py
```diff
@@ -154,4 +154,6 @@ def verify_and_update_password_hash(password, hash, salt=None):
 
 
 def hash_password(password):
-    return py.builtin._totext(pwd_context.hash(password))
+    s = py.builtin._totext(pwd_context.hash(password))
+    print('\n{} => {}\n'.format(password, s))
```
fschulze: Is this a debug leftover? I think the password shouldn't be logged.

fschulze: Sorry, I somehow looked at an outdated version

fschulze:
It would also be good if you could rebase with current master, or merge it into your branch to fix the test failures caused by newer pytest.

fschulze:
I'll work on this today

apalsson:
Thanks, and apologies for not coming back earlier to fix the issues. I hope the code will be useful for others as much as it has been to my own needs.

PR #633, opened 2019-03-13
Fix #520: submit on URL without trailing slash.

### LINKED ISSUE #520: Test upcoming Twine release (1.10.0rc1)?
https://test.pypi.org/project/twine/1.10.0rc1/  is a release candidate for Twine 1.10.0 which I'm planning to release early this week. Please see the [changelog](https://twine.readthedocs.io/en/latest/changelog.html) for detailed notes under "Next feature release".

I believe this is how you test it out on its own:  `pip install --upgrade --pre --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple twine` 

I would to hear about any problems this new release causes for `devpi`. And if there are some simple tests I can run before a new release to check whether a change in `twine` would break core functionality for you, maybe you could mention that on pypa/twine#314.

Thanks.

fschulze:


fschulze:
The test failures only occur, because PRs test against public releases and not the current commit.

PR #634, opened 2019-03-13
Client plugins

fschulze:


fschulze:
The test failures only occur, because PRs test against public releases and not the current commit.

PR #651, opened 2019-04-22
Fix #622: Skip remove if the ver comes from a base

### LINKED ISSUE #622: `devpi remove` raises 404 if inheriting from index containing the same package
## Summary

Attempting to delete a package from an index which inherits from another index (which has a package by the same name) causes an error.

This causes an issue as my workflow is to "stage" releases to a development or staging index prior to releasing them to a production index.

## Expected results

When deleting from an index, I do not expect it to list versions of the package which are not present in that index.
When deleting from an index, I only expect it to hit the `DELETE` endpoint for packages which are in that index.

## Actual results

When deleting from an index, it lists packages by the same name in parent indices.
When deleting from an index, the `DELETE` endpoint is requested even if that package version is only present in a parent index. This raises a 404 and the program exits.

## Replication

[I have created a repository that replicates the problem from scratch](https://github.com/DomHudson/devpi-bug-replication). Simply clone repository and run `script.sh`.

## Replication (details)

The steps that the script in the above repository calls are as follows:

0. Create a virtualenv and devpi instance etc

1. Login to the server.
```bash
    devpi use http://localhost:8000
    devpi login root --password=
```

2. Set up indices.
```bash
    devpi index -c production
    devpi index -c development bases=root/production
```

3. Upload the package to the production index.
```bash
    cd test_package upload
    devpi use root/production
    devpi upload
    cd ..
```

4. Upload a package with the same name (but a newer version) to the development index.
```bash
    cd test_package_v2
    devpi use root/development
    devpi upload
    cd ..
```

**At this point, we have v1.0.0 on the production index, and v2.0.0 on the development index.**

5. Try and remove the package from development index.
```bash
    devpi use root/development
    devpi remove test_package
```

The stdout out of this remove command indicates both version 1.0.0 and version 2.0.0 and claims that it will delete version 1.0.0 of the package even though the development index does not have this version.
```bash
    About to remove the following releases and distributions
    version: 1.0.0
      - No releases
    version: 2.0.0
      - http://localhost:8000/root/development/+f/734/2e0f2266bd498/test_package-2.0.0.tar.gz
    Are you sure (yes/no)?
```

Typing `yes` in response to this prompt returns an error:
```
    Are you sure (yes/no)? yes
    deleting release 1.0.0 of test-package
    DELETE http://localhost:8000/root/development/test-package/1.0.0
    404 Not Found: version '1.0.0' of project 'test-package' not found on stage 'root/development'
```

## Bypass

A way to bypass this bug is to specify the exact version number when you wish to delete from the inheriting index.

```bash
devpi remove test_package==2.0.0
```

daonb:
When removing a package, the client used to try and delete all versions found, even those residing in one of the bases.  Sorry for not writing tests, but it's really a trivial fix and if you think I have to, please point me to a sample test I can copy, preferably one with a setup that includes two indexes where one is the base of the other.

Thanks for a great tool!

Benny

### DIFF
--- a/client/devpi/list_remove.py
+++ b/client/devpi/list_remove.py
@@ -146,7 +146,8 @@ def main_remove(hub, args):
     if confirm_delete(hub, ver_to_delete):
         for ver, links in ver_to_delete:
             hub.info("deleting release %s of %s" % (ver, req.project_name))
-            hub.http_api("delete", proj_url.addpath(ver))
+            if links:
+                hub.http_api("delete", proj_url.addpath(ver))
     else:
         hub.error("not deleting anything")
 


fschulze:
You fix won't work in some cases. I created #652 instead.

Thanks for investigating and hinting me at the right track!

PR #659, opened 2019-05-09
Remove deprecated index options.

fschulze:
These were kept for compatibility with devpi-client <= 2.4.1.

The pypi_whitelist option was removed everywhere.
The acl_upload and bases options were removed from mirrors.

This doesn't require an export/import cycle.
Existing pypi_whitelist options in the db are ignored.
Upon import of old exports they will be removed.


server/devpi_server/importexport.py
```diff
@@ -343,7 +343,7 @@ def import_all(self, path):
                 user, index = stagename.split("/")
                 user = self.xom.model.get_user(user)
                 # due to possible circles we create without bases first
-                bases = indexconfig.pop('bases')
+                bases = indexconfig.pop('bases', None)
```
gotcha: Add a comment to explain BBB

server/test_devpi_server/test_model.py
```diff
@@ -506,16 +506,10 @@ def test_whitelist_setting(self, pypistage, stage, setting, expected):
...
-        assert ixconfig['pypi_whitelist'] == []
-        assert ixconfig['mirror_whitelist'] == expected
-        stage.modify(pypi_whitelist=setting)
-        ixconfig = stage.get()
-        assert ixconfig['pypi_whitelist'] == []
-        assert ixconfig['mirror_whitelist'] == expected
-        stage.modify(pypi_whitelist=[], mirror_whitelist=setting)
-        ixconfig = stage.get()
-        assert ixconfig['pypi_whitelist'] == []
+        assert 'pypi_whitelist' not in ixconfig
```
gotcha: Add a comment explaining this is checking for BBB

Issue #658, opened 2019-05-09
Always require --storage option for custom backends


fschulze:
Currently the ``--storage`` option is only required when initializing a devpi-server instance with ``--init``. It writes those options into ``.nodeinfo`` and looks them up when ``--storage`` isn't given.

This should be changed, so ``--storage`` is always required.

For the default case of using the sqlite backend with file storage, we should detect whether the ``.sqlite`` file exists. If it doesn't, then we tell the user that the ``--serverdir`` doesn't contain the sqlite database and remind him of the ``--storage`` option or the ``--init`` option in case the didn't initialize devpi-server yet.

This is a pre-requisit to get rid of ``.nodeinfo``.

fschulze:
Also see ``_determine_storage`` in ``config.py``.

fschulze:
Will be in devpi-server 5.0.0

PR #660, opened 2019-05-09
require --storage option

gotcha:
#658  Do not persist storage in nodeinfo.


gotcha:
Test missing for Fatal error when .sqlite file is missing

PR #664, opened 2019-05-16
Add support for index customizers

fschulze:


server/devpi_server/model.py
```diff
@@ -149,6 +149,15 @@ def getstage(self, user, index=None):
             return user.getstage(index)
 
 
+def ensure_boolean(value):
+    if isinstance(value, bool):
```
gotcha: check if we can make treatment of yes or true symetric to no or false

server/devpi_server/model.py
```diff
@@ -338,17 +338,17 @@ def create_stage(self, index, type="stage", **kwargs):
...
+            kwargs.setdefault("mirror_whitelist", [])
         ixconfig = get_indexconfig(
             self.xom.config.hook,
             type=type, **kwargs)
-        if type == "stage":
-            if ixconfig.get("acl_upload") is None:
-                ixconfig["acl_upload"] = [self.name]
-            if ixconfig.get("acl_toxresult_upload") is None:
-                ixconfig["acl_toxresult_upload"] = [":ANONYMOUS:"]
+        if "bases" in ixconfig:
```
gotcha: might need to be moved to the ``if type == "stage"`` above

server/devpi_server/model.py
```diff
@@ -343,35 +345,27 @@ def create_stage(self, index, type="stage", **kwargs):
...
-                raise InvalidIndex("indexname '%s' already exists" % index)
-            indexes[index] = ixconfig
-        stage = self.getstage(index)
+        stage = self._getstage(index, type, {"type": type})
+        stage._modify(**kwargs)
         threadlog.info("created index %s: %s", stage.name, stage.ixconfig)
         return stage
 
+    def _getstage(self, indexname, index_type, ixconfig):
+        if index_type == "mirror":
```
gotcha: I do not like branching on ``index_type``

server/pytest_devpi_server/__init__.py
```diff
@@ -16,5 +16,5 @@ def devpiserver_makepypistage():
     def makepypistage(xom):
         from devpi_server.main import _pypi_ixconfig_default
         return PyPIStage(xom, username="root", index="pypi",
-                         ixconfig=_pypi_ixconfig_default)
+                         ixconfig=dict(_pypi_ixconfig_default))
```
gotcha: comment why doing a copy of dict is useful

server/devpi_server/extpypi.py
```diff
@@ -153,6 +153,19 @@ def mirror_url(self):
             url = URL(self.ixconfig['mirror_url'])
             return url.asdir().url
 
+    def _iter_indexconfig_items(self, **kwargs):
```
gotcha: rename to ``indexconfig_items``

server/devpi_server/model.py
fschulze: Changed in https://github.com/devpi/devpi/pull/664/commits/a94d47ad323d7a8a81945da093b0a53ecf88dabf

server/pytest_devpi_server/__init__.py
fschulze: Add in 0a59c7f6

server/devpi_server/extpypi.py
fschulze: Done in 4b88b514

server/test_devpi_server/test_model.py
```diff
@@ -1028,7 +1028,7 @@ def test_get_indexconfig_lists(xom, key, value, result):
      dict(type="stage")),
 
     ({"volatile": "foo"},
-     dict(type="stage", volatile=True)),
+     InvalidIndexconfig),
```
gotcha: Tests with ``"True"`` or ``"true"`` are missing

fschulze: fixed in 5580a5be

server/devpi_server/views.py
```diff
@@ -667,7 +667,9 @@ def index_modify(self):
             json = ixconfig
         kvdict = getkvdict_index(self.xom, json)
         try:
+            oldconfig = dict(stage.ixconfig)
             ixconfig = stage.modify(**kvdict)
+            stage.customizer.on_modified(self.request, oldconfig)
         except InvalidIndexconfig as e:
```
gotcha: Can you catch ``HTTPException`` to let the plugin developer lnow that he should use ``apifatal`` ?

server/test_devpi_server/test_stage_customizer.py
```diff
@@ -157,3 +157,25 @@ def validate_config(self, oldconfig, newconfig):
     assert r.json['message'] == "You can't have fewer bases than before."
     result = mapp.getjson(api.index)
     assert result['result']['bases'] == ['user/foo']
+
+
```
gotcha: a test is missing for succes case of ``on_modified``

server/devpi_server/model.py
```diff
@@ -311,6 +317,8 @@ def create_stage(self, index, type="stage", **kwargs):
         if index in self.get().get("indexes", {}):
             raise InvalidIndex("indexname '%s' already exists" % index)
         stage = self._getstage(index, type, {"type": type})
+        if isinstance(stage.customizer, UnknownCustomizer):
+            raise InvalidIndexconfig("unknown index type %r" % type)
```
gotcha: I see no tests for this ``InvalidIndexconfig``.

server/devpi_server/model.py
```diff
@@ -499,6 +525,8 @@ def get_link_from_entrypath(self, entrypath):
         return links[0] if links else None
 
     def store_toxresult(self, link, toxresultdata):
+        if self.customizer.readonly:
+            raise ReadonlyIndex("index is marked read only")
```
gotcha: I see no tests for ``ReadonlyIndex`` exceptions

server/devpi_server/views.py
```diff
@@ -239,6 +241,13 @@ def get_actions(json):
     return result
 
 
+@exception_view_config(ReadonlyIndex)
```
gotcha: I see no tests for the ``readonly_index_view``.

server/devpi_server/model.py
```diff
@@ -359,6 +371,32 @@ def verify_name(cls, indexname):
...
+        return self.stage.ixconfig.get("acl_toxresult_upload", [':ANONYMOUS:'])
+
+    def get_index_delete_principals(self, restrict_modify=None):
+        if restrict_modify is None:
+            modify_principals = set(['root', self.stage.username])
+        else:
+            modify_principals = restrict_modify
+        return modify_principals
+
+    get_index_modify_principals = get_index_delete_principals
```
gotcha: Why is ``modify`` same as ``delete``. There should be a comment.

gotcha: Ok it is not that delete is same as modify. It is that they share permissions.

gotcha: Naming of those methods is maybe not obvious. I'd rather go for ``get_principals_for_index_delete``.

server/devpi_server/model.py
fschulze: In this case it is not avoidable (yet). But the final version of this PR removed several similar cases.

server/devpi_server/model.py
fschulze: It is gone  now with "Split index type specific parts of get_indexconfig into separate methods."

server/devpi_server/views.py
fschulze: This is fixed with "Add on_modified method for stage customizer plugins."

server/test_devpi_server/test_stage_customizer.py
fschulze: fixed

server/devpi_server/model.py
fschulze: Covered by test_create_with_invalid_type

server/devpi_server/model.py
fschulze: Changed

server/devpi_server/model.py
fschulze: These are not reached by regular code, because the permissions for UnknownCustomizer shortcut it. These exceptions are thus an additional security net. Not sure whether we should explicitly test for them.

server/devpi_server/views.py
fschulze: See explanation for the exceptions.

server/devpi_server/model.py
fschulze: The test which would reach these if the permissions wouldn't work is ``test_permissions_for_unknown_index``

PR #674, opened 2019-05-31
Allow mirrors to serve external urls

fschulze:


fschulze:
Only files use external URLs for now, the project data is still stored in the DB.

RonnyPfannschmidt:
looks interesting, i'll take a feeler look on monday

fschulze:
I simplified this quite a bit. When the ``mirror_use_external_urls`` option is ``True`` then any package not existing locally will return a redirect to the original URL.

PR #675, opened 2019-06-02
Use ruamel.yaml for --logger-cfg.

fschulze:


StephanErb:
@fschulze I have verified that we can load our existing configuration. Thanks for the heads-up!

PR #681, opened 2019-06-26
Docs: More details on using YAML configuration file.

smsearcy:
I had to look through some source code to figure out the specifics on the file and application names used with *appdirs* in order to use a YAML configuration file.  It seemed like the docs would benefit from having that information spelled out so I'm submitting this PR.

### DIFF
--- a/doc/quickstart-server.rst
+++ b/doc/quickstart-server.rst
@@ -44,7 +44,12 @@ Using a configuration file for devpi-server
 
 A `strict YAML`_ conform configuration file can be used in place of the command line options.
 
-The configuration file will be looked for in your systems and your users default configuration location using the `appdirs`_ library.
+A configuration file named ``devpi-server.yml`` will be looked for in your systems and your users default configuration location using the `appdirs`_ library.
+
+.. note::
+
+    For ``appdirs`` the application name is ``devpi-server``, and the application author is ``devpi``.
+    Thus, for example, on Linux the user configuration file would be ``~/.config/devpi-server/devpi-server.yml``.
 
 With the `-c/--configfile` option the location can be provided explicitly.
 


fschulze:
Thanks!

doc/quickstart-server.rst
```diff
@@ -44,7 +44,12 @@ Using a configuration file for devpi-server
 
 A `strict YAML`_ conform configuration file can be used in place of the command line options.
 
-The configuration file will be looked for in your systems and your users default configuration location using the `appdirs`_ library.
+A configuration file named ``devpi-server.yml`` will be looked for in your systems and your users default configuration location using the `appdirs`_ library.
+
+.. note::
+
+    For ``appdirs`` the application name is ``devpi-server``, and the application author is ``devpi``.
+    Thus, for example, on Linux the user configuration file would be ``~/.config/devpi-server/devpi-server.yml``.
```
gotcha: @fschulze What would it be for Windows ? mentioning the application author when it seems to not be used in the Linux example is confusing

smsearcy: Good question, on Windows the application author is used in the path, that's why I included it but including example paths for Windows would have been a good idea.

I haven't tested this but based on the *appdirs* documentation I think on a typical Windows 7+ setup for user config it would be ``C:\Users\<username>\AppData\Local\devpi\devpi-server\devpi-server.yml`` and the system configuration would be ``C:\ProgramData\devpi\devpi-server\devpi-server.yml``.

fschulze: I would say we should refer to the appdirs docs and code. If someone tests an example, we can add it. I might be able to try a basic Windows 10 Home installation.

PR #701, opened 2019-08-27
Common package and improve setup file

krmanish:
~~Improve setup script to create rpm or build spec.~~


common/setup.py
```diff
@@ -25,18 +25,26 @@ def get_changelog():
     setup(
       name="devpi-common",
       description="utilities jointly used by devpi-server and devpi-client",
-      long_description="\n\n".join([README, CHANGELOG]),
```
fschulze: Why did you remove the changelog? We have it here explicitly, so people can see the last changes easily from the pypi page.

common/setup.py
```diff
@@ -25,18 +25,26 @@ def get_changelog():
...
       maintainer_email="holger@merlinux.eu",
+      license='MIT',
       classifiers=[
         "Development Status :: 5 - Production/Stable",
         "Environment :: Web Environment",
         "Intended Audience :: Developers",
         "Intended Audience :: System Administrators",
         "License :: OSI Approved :: MIT License",
+        "Programming Language :: Python",
+        "Programming Language :: Python :: 2.6",
```
fschulze: The Python classifiers you added don't reflect the current state. There are long obsolete versions and others like 3.6 are missing.

common/setup.py
krmanish: After installing the package using `yum`, I used to see the rpm spec. Imo having change log in the rpm spec make no sense. `rpm -qi devpi-common`.

common/setup.py
krmanish: > I don't see how these changes have anything to do with building rpms. If you could explain how these help, we can find a way to fix this properly.

Yes, It's my bad. Building package or creating rpm has nothing to do with it. However, I think improvement is needed in the rpm spec.

krmanish: > The Python classifiers you added don't reflect the current state. There are long obsolete versions and others like 3.6 are missing.

Agree, I didn't go through the entire log to see all supported version, Yes will include 3.6 for sure

fschulze:
I fixed this differently on master, could you give it a try?

PR #733, opened 2019-10-26
Fix #732: add --unix-socket-perms option

### LINKED ISSUE #732: `unix-socket` option needs companion `unix-socket-perms` option
When running in a Docker container (for example), Nginx might be running as user `www-data`, but Devpi might be running as some other user. If you use the new `--unix-socket` option, Waitress creates the the Unix socket to be owned by the user running Devpi with default permissions `600`, which means that no other users can use the socket. I'm currently working around this by spinning up something to `chmod` the file after Devpi starts, but this is not ideal. Waitress has a `unix_socket_perms` option that defaults to `'600'` (the string, not the number). Devpi should add a `--unix-socket-perms` option to mirror it. :-) 

See: https://docs.pylonsproject.org/projects/waitress/en/stable/api.html

Related to: issue #253, PR #718

beamerblvd:


fschulze:
Thanks!

PR #734, opened 2019-10-28
Build wheels compatible with python2 and python3

favipcj:


client/devpi/upload.py
```diff
@@ -414,6 +414,9 @@ def setup_build(self, default_formats=None):
             pre = [self.python, "setup.py"]
             cmd = pre + buildcommand
 
+            if format == "bdist_wheel":
+                cmd = cmd + ["--universal"]
```
RonnyPfannschmidt: That's fundamental wrong for binary shipping packages

fschulze:
To get universal wheels for your packages, use a setup.cfg file like this of devpi-client itself:
https://github.com/devpi/devpi/blob/master/client/setup.cfg

PR #739, opened 2019-11-08
Add ssl options to devpi-postgresql

### LINKED ISSUE #738: postgres storage plugin can't use SSL
The postgres storage plugin exposes most useful connection options from pg8000, but not "ssl".

JonathanRRogers:
Fixes #738 
Add ``ssl_cert_reqs``, ``ssl_ca_certs``, ``ssl_certfile`` and ``ssl_keyfile`` settings to the --storage=pg8000 option. Each setting corresponds to the similarly-named key (remove the initial "ssl_") in the ``ssl`` parameter passed to pg8000.connect(). ``ssl_cert_reqs`` should be either "cert_optional" or "cert_required" and the other three settings are file names.


fschulze:
Is it necessary to do SSL for all the tests? I'm pretty sure the test run time decreases due to that (haven't benchmarked yet), because we recreate the db and connection for almost every test. A single test explicitly testing the ssl options seems better to me.

The tox.ini would also need the certauth package for the tests to work everywhere.

A changelog entry would be nice, see https://github.com/devpi/devpi/blob/master/CONTRIBUTING.rst#pull-requests and https://github.com/devpi/devpi/blob/master/server/news/examples.rst

fschulze:
Besides that this looks good!

JonathanRRogers:
> Is it necessary to do SSL for all the tests? I'm pretty sure the test run time decreases due to that (haven't benchmarked yet), because we recreate the db and connection for almost every test. A single test explicitly testing the ssl options seems better to me.

OK, I'll add add a test just for the SSL connection.

> 
> The tox.ini would also need the certauth package for the tests to work everywhere.

OK, I'll add it to tox.ini as well as to "dev-bootstrap.sh". I don't know if I can make the test work on Windows.

> 
> A changelog entry would be nice, see https://github.com/devpi/devpi/blob/master/CONTRIBUTING.rst#pull-requests and https://github.com/devpi/devpi/blob/master/server/news/examples.rst

I did add a changelog entry in "postgresql/news/738.feature". Should it go somewhere else?

JonathanRRogers:
I didn't find an easy way to add a test specifically for SSL connections because current tests work by monkey-patching devpi_postgresql.main. Instead, I added a tox environment which runs one of the existing server tests with a new command line parameter. I don't think this approach saved much time vs. using SSL connections for all PostgreSQL tests, but covering both configurations is better anyway. It seems one of the Windows tests failed, but the failure doesn't seem to have anything to do with PostgreSQL.

fschulze:
Thanks!

PR #741, opened 2019-11-22
fixing devpi web error when info about more than 10 replication error

r2dan:
Hi Florian,

I added a quick fix to one issue in devpi-web I noticed recently - when number of replication errors is more than 10, then for each error with index greater than 10 an extra message "More than 10 replication errors" is added. This fix makes sure it is added only once.

Unfortunately I did not find any unit tests covering replication errors. I can add one if needed - please let me know how to handle it. Thanks!

### DIFF
--- a/web/devpi_web/views.py
+++ b/web/devpi_web/views.py
@@ -729,8 +729,8 @@ def statusview(request):
     replication_errors = []
     for index, error in enumerate(status.get('replication-errors', {}).values()):
         replication_errors.append(error)
-        if index >= 10:
-            replication_errors.append(dict(message="More than 10 replication errors."))
+    if len(replication_errors) >= 10:
+        replication_errors.append(dict(message="More than 10 replication errors."))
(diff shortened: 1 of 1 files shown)


fschulze:
The intent was to abort the list after 10 items, so I think this would be a better fix:

```diff
diff --git a/web/devpi_web/views.py b/web/devpi_web/views.py
index 4b83f8f0..59dc03ce 100644
--- a/web/devpi_web/views.py
+++ b/web/devpi_web/views.py
@@ -731,6 +731,7 @@ def statusview(request):
         replication_errors.append(error)
         if index >= 10:
             replication_errors.append(dict(message="More than 10 replication errors."))
+            break
     _polling_replicas = status.get('polling_replicas', {})
     polling_replicas = []
     for replica_uuid in sorted(_polling_replicas):
```

The replication is being worked on in a branch and this will be obsolete, so I don't think we need to bother about a test.

fschulze:
This is fixed on master.

PR #744, opened 2019-11-30
Proxy timeout

fschulze:
cc @r2dan


r2dan:
It's great to see you increased it a little bit. Thanks!

Issue #746, opened 2019-12-03
remove devpi-client's tox dependency for non-development install


cjw296:
I'm confused why tox is a dependency of devpy-client. It seems unrelated to devpi-client's use case and is a pretty heavy dependency that [causes ](https://github.com/tox-dev/tox/pull/1472) [issues](https://github.com/tox-dev/tox/issues/1473).

Would it be possible to remove 'tox' and hopefully 'check-manifest' and maybe even 'pkginfo', or at least make them development-only dependencies?

fschulze:
We might be able to get rid of the tox dependency, since detox is obsolete now and we could just use subprocess now.

The check-manifest and pkginfo packages are used by the upload command and are essential to its functionality. It would be very confusing for users if they would have to install these manually to be able to use ``devpi upload``.

fschulze:
Oh, forgot to mention that ``devpi test`` requires tox.

wimglenn:
Perhaps it could be made into an optional dependency, like `devpi-client[test]`.  I have no interest to use `devpi test` but I do want the twine-like stuff  `use`, `login`, `upload`...

fschulze:
@wimglenn are you aware, that for uploads you can use twine directly? Maybe that is a better solution for you? For other devpi maintenance you would still need devpi-client. We may also be extending the devpi-web UI for some basics in the future.

wimglenn:
Yes, I am aware that twine works for upload, but the devpi-client has several other useful features too..

cjw296:
@fschulze - can twine upload docs to devpi? If so, how? If not, are there any other tools beside devpi-client that can?

fschulze:
I haven't tried twine for docs, but my suspicion is that it doesn't, because PyPI doesn't support that for quite some time now, but the protocol is almost the same, so it might work.

Using ``python setup.py ...`` might also still work.

fschulze:
No reply in more than 3 months. Feel free to reopen this issue if it isn't solved yet.

cjw296:
@fschulze - not sure what reply you were waiting for but nothing has changed and this issue is still a problem. I think @wimglenn's suggestion for `devpi-client[test]`, given that `tox` is pretty huge and onerous dependency.

It looks like the PR is stalled on @wimglenn making the changes you requested, not sure if it makes sense to just track there or to re-open this issue?

fschulze:
Feel free to make a new PR with the required changes. I'm happy to accept it as described, but won't do the work myself.

[...]

PR #747, opened 2019-12-03
Remove test stuff from devpi-client's packaging

wimglenn:
Closes https://github.com/devpi/devpi/issues/746 

- Make the tox dependency optional.  Can use `pip install devpi-client[test]` to pull it in.
- Don't export a `testing` package to site-packages on `pip install devpi-client` 

```
$ ls ~/.venv/lib/python3.8/site-packages/testing
conftest.py    __init__.py  reqmock.py  test_functional.py  test_install.py      test_login.py  test_push.py    test_test.py    test_use.py
functional.py  __pycache__  simpypi.py  test_index.py       test_list_remove.py  test_main.py   test_pypirc.py  test_upload.py  test_user.py
```

### DIFF
--- a/client/setup.py
+++ b/client/setup.py
@@ -26,14 +26,15 @@ if __name__ == "__main__":
     README = io.open(os.path.join(here, 'README.rst'), encoding='utf-8').read()
     CHANGELOG = get_changelog()
 
-    install_requires=["tox>=3.1.0",
-                      "devpi_common<4,>=3.4.0",
+    install_requires=["devpi_common<4,>=3.4.0",
                       "pkginfo>=1.4.2",
                       "check-manifest>=0.28",
                       "pluggy>=0.6.0,<1.0",
                       "py>=1.4.31"]
 
-    extras_require = {}
+    extras_require = {
+        "test": ["tox>=3.1.0"],
+    }
 
     setup(
       name="devpi-client",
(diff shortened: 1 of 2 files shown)


wimglenn:
@fschulze   Could you take a look at the appveyor build please?  I don't really understand the problem, that fails somewhere in `devpi-server.exe --version` but for some reason it's using a Python-3.4 runtime in the py27 test, which seems pretty weird. Configuration problem?

fschulze:
As it is now the "devpi test" command fails with an unhelpful import error on a clean installation without ``[test]``. We are currently in contact with companies that pay for this feature to discuss possible solutions. Until that is cleared up I won't accept this PR.

I will investigate whether removing the ``testing`` package causes no other issues, but I think that part is fine.

You can ignore the AppVeyor failure, it is unrelated and I will look into it. The reason Python 3.x is used in the Python 2.7 tests is that the functional tests run devpi-server in a virtualenv and because it will soon drop Python 2.7 support it uses Python 3.x to run it.

wimglenn:
@fschulze  How about to add a message in `devpi test` which tells user they need to install extra `devpi-client[test]` to use that feature?  Like a message on stderr and then exit 1.

fschulze:
@wimglenn that is the consensus we got from asking about it. When ``devpi test`` is run, it should try to find a ``tox`` executable in the path and if that fails try to import it. If nothing can be found something along this should be displayed: ``No tox executable or module could be found. Use "pip install tox" or some other way to ensure tox is available as an executable or importable by devpi-client``.

This would also be a breaking change requiring a major version bump and proper ``news`` entry.

fschulze:
Checking for the ``tox`` executable would be a new feature, so that would be a nice to have, but the other requirements stand for accepting this PR.

wimglenn:
Replaced with https://github.com/devpi/devpi/pull/890

PR #748, opened 2019-12-05
Fix #688: on file upload existing metadata is only updated, not replaced.

### LINKED ISSUE #688: requires_dist metadata missing
When uploading a new release with `devpi upload`, I have noticed that `requires_dist` metadata is missing. Calling `devpi upload` a second time on the same source is adding the `requires_dist` metadata even on non-volatile index.

fschulze:
cc @r2dan


r2dan:
When do you plan to release a bugfix version?

fschulze:
@r2dan out now

PR #755, opened 2019-12-16
Fix caching of Whoosh indexer, so memory is properly freed.

fschulze:


web/devpi_web/whoosh_index.py
```diff
@@ -518,8 +518,9 @@ def project_schema(self):
     def delete_projects(self, projects):
         counter = itertools.count()
         count = next(counter)
-        writer = self.project_ix.writer()
-        searcher = self.project_ix.searcher()
+        project_ix = self.project_ix
```
RonnyPfannschmidt: shouldn't this be a normal method based on what it does? 
leaving it a property would make it easy to misinterpret later

fschulze: Makes sense: https://github.com/devpi/devpi/commit/27bda19aa8b91fc8a8d275eb92d148270418b8f8

RonnyPfannschmidt: lovely, thanks

PR #757, opened 2019-12-16
Fix #756: correct URL to toxresults when using latest or stable as version in URL.

### LINKED ISSUE #756: Links to toxresults broken when using "latest" or "stable" version.
The version page generates broken links to the tox results when using ``latest`` or ``stable`` as version in the URL.

fschulze:
cc @r2dan

### DIFF
--- a/web/devpi_web/views.py
+++ b/web/devpi_web/views.py
@@ -667,6 +667,7 @@ def version_get(context, request):
         title="%s/: %s-%s metadata and description" % (stage.name, name, version),
         content=get_description(stage, name, version),
         summary=verdata.get("summary"),
+        resolved_version=version,
         nav_links=nav_links,
         infos=infos,
         metadata_list_fields=frozenset(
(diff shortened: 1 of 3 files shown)


web/devpi_web/views.py
```diff
@@ -673,11 +678,11 @@ def version_get(context, request):
         make_toxresults_url=functools.partial(
             request.route_url, "toxresults",
             user=context.username, index=context.index,
-            project=context.project, version=context.version),
+            project=context.project, version=version),
         make_toxresult_url=functools.partial(
             request.route_url, "toxresult",
             user=context.username, index=context.index,
-            project=context.project, version=context.version))
+            project=context.project, version=version))
```
r2dan: Can you please return version=version here as well? It would help me in handling latest/stable versions in my custom theme.

fschulze: I wrongly assumed it would be available via the request already. I now explicitly added it to the result and added the wrapped context as ``_context`` to the other views.

r2dan:
@fschulze: when do you plan to release this fix?

PR #763, opened 2020-01-09
Add issue template for GitHub.

fschulze:


hellp:
Hey, as you asked on Twitter… Yeah, I could see this coming over as slightly passive-aggressive. Here's my variant, a bit more wordy, some more verbs, and I replaced "explain yourself" with "please note".

---

Thanks for filing an issue! Please check the following points before submitting:

- Have you tried if the lastest released version fixes your issue? If not, please note what prevents you from doing so.
- Provide the output of `pip list` from the virtual environment you are using.
- Provide the Python and operating system versions under which the issue occurs.
- If possible, provide a minimal example to reproduce the issue.

RonnyPfannschmidt:
@hellp lovely spin :+1:

PR #784, opened 2020-03-25
Upgrade pg8000 to version 1.14.0 or newer.

### LINKED ISSUE #783: pg8000 ssl interface changed
@JonathanRRogers  pg8000 1.14.0 has changed its ssl interface and I'm not sure how to adapt to it. Instead of options for wrap_socket it now wants an SSLContext object.

JonathanRRogers:
Fixes  #783


fschulze:
Thanks!

As far as I can see none of the option change, so I think a 1.1.1 is fine, maybe 1.2.0.

Why did you remove Python 3.4? Did you get test failures because of it? It seems like you branched from an old version, you should rebase from current master to get several fixes for current pytest etc.

If you don't have time, then I can do the rebase and revert the Python 3.4 removal.

postgresql/devpi_postgresql/main.py
```diff
@@ -11,6 +11,7 @@
 import pg8000
 import py
 import time
+from distutils.util import strtobool
```
fschulze: ``devpi_server.model`` has ``ensure_boolean``, I think it would be more consistent to use it, as everything else in devpi uses it.

postgresql/devpi_postgresql/main.py
```diff
@@ -183,17 +185,17 @@ def __init__(self, basedir, notify_on_commit, cache_size, settings=None):
...
+                cafile=settings.get('ssl_ca_certs'))
 
-            for key in ("ssl_ca_certs", "ssl_certfile", "ssl_keyfile"):
-                if key in settings:
-                    ssl_opts[key[4:]] = settings[key]
+            if 'ssl_certfile' in settings:
+                ssl_context.load_cert_chain(settings['ssl_certfile'],
+                                            keyfile=settings.get('ssl_keyfile'))
+
+            check_hostname = settings.get('ssl_check_hostname')
```
fschulze: This isn't in SSL_OPT_KEYS, is that an oversight?

fschulze:
Ok, looks like ``ssl_cert_reqs`` is gone and ``ssl_check_hostname`` is new, that change would certainly justify a major version bump.

postgresql/devpi_postgresql/main.py
JonathanRRogers: > Ok, looks like `ssl_cert_reqs` is gone and `ssl_check_hostname` is new, that change would certainly justify a major version bump.

Yes, I think I made a mistake to include `ssl_cert_reqs` in the first place because I didn't quite understand what it did. Based on my current reading of the standard library documentation, either value ("cert_optional" or "cert_required") would have had the same effect.

The addition of `ssl_check_hostname` is the primary purpose of this pull request. In contrast with my earlier pull request, secure connections will now default to checking host name. It may be desirable to disable this.

JonathanRRogers: > This isn't in SSL_OPT_KEYS, is that an oversight?

The ssl settings are now "ssl_check_hostname", "ssl_ca_certs", "ssl_certfile" and "ssl_keyfile" and they are all in SSL_OPT_KEYS. Did I miss something?

JonathanRRogers:
> Why did you remove Python 3.4? Did you get test failures because of it? It seems like you branched from an old version, you should rebase from current master to get several fixes for current pytest etc.

Since version 1.14.0, pg8000 requires Python 3.5+. AFAICT,  JonathanRRogers/devpi has one commit (eec0a6a7150367e26452ecde92f4d02614cd4341) more recent than the most recent in devpi/devpi (6ea708ebf558b9032d8934b16c50dcb16d0b162f).

postgresql/devpi_postgresql/main.py
JonathanRRogers: OK, it now uses ensure_boolean.

JonathanRRogers:
The most recent Travis build passed except for the nightly client, which also failed in the most recent master build.

fschulze:
I have cherry picked the first and last commit and skipped the pytest pinning. The test failures only happen on PRs, because they use PyPI releases without the fixes from master.
Thanks!

JonathanRRogers:
> I have cherry picked the first and last commit and skipped the pytest pinning. The test failures only happen on PRs, because they use PyPI releases without the fixes from master.
> Thanks!

Yes, I wasn't sure about the best way to deal with the tests failing due to incompatibilities between devpi-server 5.4.0 and pytest 5.x. Perhaps the Travis build should always install devpi-server from the git repository? Thanks.

fschulze:
I haven't found a solution yet to do that in a way that works with travis, locally and ``devpi test``.

PR #821, opened 2020-10-13
Fix tests: do not rely on os.listdir order

jfroche:
os.listdir returns an arbitrary ordered list of files [1]

Here is an example of failing tests due to this non ordered: https://hydra.nixos.org/build/127650244/nixlog/2

[1] https://docs.python.org/3/library/os.html#os.listdir


fschulze:
Thanks!

PR #824, opened 2020-11-15
Apply "pkg_read" permissions

andrenerd:
Hello,

please, add permissions for reading packages (aka "projects", if i get it right).
it's really needed to set user-based access to packages further (via plugins).

ps. if you find "pkg_read" is not ok, lets call it "read_project" (based on another existing permission name  -"del_project") or so.

### DIFF
--- a/server/devpi_server/views.py
+++ b/server/devpi_server/views.py
@@ -533,7 +533,7 @@ class PyPIView:
             index=self.context.index,
             project=self.context.project))
 
-    @view_config(route_name="/{user}/{index}/+simple/{project}/")
+    @view_config(route_name="/{user}/{index}/+simple/{project}/", permission="pkg_read")
     def simple_list_project(self):
         request = self.request
         abort_if_invalid_project(request, request.matchdict["project"])
@@ -1114,7 +1114,7 @@ class PyPIView:
             project=self.context.project))
 
     @view_config(route_name="/{user}/{index}/{project}",
-                 accept="application/json", request_method="GET")
+                 accept="application/json", request_method="GET", permission="pkg_read")
     def project_get(self):
         if not json_preferred(self.request):
             apireturn(415, "unsupported media type %s" %


fschulze:
There is something in the works for this already. Your suggestion gave me a good idea for simplification of that. Thanks! I'll keep this open until it lands on master.

ranxian:
@fschulze Wondering if there is still ongoing work for this feature and what's the roadmap, it would be greatly useful for our workflow

fschulze:
The server side is in the latest devpi-server and devpi-lockdown releases, but the client side still needs further testing and a release. One also needs to write a plugin with custom hook implementations which I will most likely put in a gist as an example. With that a whole index can be restricted, individual projects have too many interactions with other parts to be easily restricted without unwanted side effects or leaks.

fschulze:
The client side changes have been merged for the upcoming devpi-client 6.0.0.

The hooks could look like this:

```
from pluggy import HookimplMarker


server_hookimpl = HookimplMarker("devpiserver")


@server_hookimpl
def devpiserver_authcheck_forbidden(request):
    if request.authenticated_userid:
        stage = request.context._stage
        if stage and not request.has_permission("pkg_read"):
            return True


@server_hookimpl
def devpiserver_indexconfig_defaults(index_type):
    return {"acl_pkg_read": ACLList([":ANONYMOUS:"])}


@server_hookimpl
def devpiserver_stage_get_principals_for_pkg_read(ixconfig):
    return ixconfig.get("acl_pkg_read")
```

If you want the lockdown to spread to inheriting indexes, add this hook as well:

```
@server_hookimpl
def devpiserver_sro_skip(stage, base_stage):
    request = get_current_request()
    if request is not None and not request.has_permission("pkg_read", base_stage):
        return True
```

As this still isn't a perfect lockdown I won't add this to an existing plugin or create a new one. Everyone has to test this for their own circumstances. For example the devpi-web search can leak names and there is most likely other stuff I currently forget.

Issue #835, opened 2021-03-01
[5.5.0][pyramid=2.0] ModuleNotFoundError: No module named 'pyramid.compat'


hexagonrecursion:
When I run `devpi-server --version` I get `ModuleNotFoundError: No module named 'pyramid.compat'`. 

- [x] Are you using the latest released version? (5.5.0)
- [x] Provide the output of ``pip list`` from the virtual environment you are using.
  <details>
    <summary>pip list</summary>

  ```
  Package             Version
  ------------------- ---------
  apipkg              1.5
  appdirs             1.4.4
  argon2-cffi         20.1.0
  attrs               20.3.0
  certifi             2020.12.5
  cffi                1.14.5
  chardet             4.0.0
  devpi-common        3.6.0
  devpi-server        5.5.0
  execnet             1.8.0
  hupper              1.10.2
  idna                2.10
  itsdangerous        1.1.0
  lazy                1.4
  passlib             1.7.4
  PasteDeploy         2.1.1
  pip                 21.0.1
  plaster             1.0
  plaster-pastedeploy 0.7
  pluggy              0.13.1
  py                  1.10.0
  pycparser           2.20
  pyramid             2.0
  python-dateutil     2.8.1
  repoze.lru          0.7
  requests            2.25.1
  ruamel.yaml         0.16.12
  setuptools          49.2.1
  six                 1.15.0
  strictyaml          1.3.2
  translationstring   1.4
  urllib3             1.26.3
  venusian            3.0.0
  waitress            1.4.4
  WebOb               1.8.7
  zope.deprecation    4.4.0
  zope.interface      5.2.0
  ```

  </details>
- [x] Provide the Python and operating system versions under which the issue occurs.
  ```
  Python 3.9.1
  Fedora release 32 (Thirty Two)
  ```
- [x] If possible, provide a minimal example to reproduce the issue.
  ```
   $ devpi-server --version
  Traceback (most recent call last):
    File "/path/to/venv/bin/devpi-server", line 5, in <module>
      from devpi_server.main import main
    File "/path/to/venv/lib/python3.9/site-packages/devpi_server/main.py", line 18, in <module>
      from .views import apireturn
    File "/path/to/venv/lib/python3.9/site-packages/devpi_server/views.py", line 18, in <module>
      from pyramid.compat import escape
  ModuleNotFoundError: No module named 'pyramid.compat'
  ```

Could it be that `pyramid.compat` was removed in `pyramid=2.0`?
[...]

PR #836, opened 2021-03-01
Cap dependency range: pyramid<2

hexagonrecursion:
Fixes https://github.com/devpi/devpi/issues/835

### DIFF
--- a/server/setup.py
+++ b/server/setup.py
@@ -32,7 +32,7 @@ if __name__ == "__main__":
                         "devpi_common<4,>=3.3.0",
                         "itsdangerous>=0.24",
                         "execnet>=1.2",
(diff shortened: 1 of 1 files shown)


fschulze:
Thanks, but unless 6.0.0 will take a lot longer I won't make another release of the 5.x line. I hope to make the release by the end of this week.

gaborbernat:
@fschulze IMHO this should be done in a small patch release ``5.5.1``.  This is a devpi issue, and not need to be addressed in downstream tools (see https://github.com/tox-dev/tox/pull/1930 opened).

fschulze:
I'll do a 5.5.1, since this affects other test runs as well.

PR #847, opened 2021-09-02
📚 Fix docs typo

Ambro17:
- Fix Minor typo 
- Update footnotes removing old reference that no longer existed since 03dd0de


fschulze:
Thanks!

PR #857, opened 2021-10-12
Add missing dependency to devpi-server

spire-allyjweir:
`lazy` is used within devpi-server but was depending on it as a
transitive dependency from the devpi-common>=3.5.0 package being
available.

* [Example import of `lazy` in devpi-server v6.2.0](https://github.com/devpi/devpi/blob/main/server/devpi_server/views.py#L18)


fschulze:
I think an explicit dependency is better.

spire-allyjweir:
Happy to fix this up and will add a changelog record. Thanks for the guidance @fschulze

PR #875, opened 2022-01-25
Update launchd-macos.txt.template

jquade:
Updated the template text to match the net.devpi.plist file generated by server/devpi_server/genconfig.py.

Removed the “start” invocation, because RunAtLoad is used in the generated config file.


fschulze:
Thanks!

PR #890, opened 2022-05-09
remove devpi-client's tox dependency

wimglenn:
Replaces https://github.com/devpi/devpi/pull/747 
Closes https://github.com/devpi/devpi/issues/746

### DIFF
--- a/client/devpi/main.py
+++ b/client/devpi/main.py
@@ -1,5 +1,6 @@
 # PYTHON_ARGCOMPLETE_OK
 import os
+import shutil
 import sys
 import py
 import argparse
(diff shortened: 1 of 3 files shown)


fschulze:
Thanks for the PR. This is already in the branch for the upcoming devpi-client 6.x in my fork:
https://github.com/fschulze/devpi/commits/client6
https://github.com/fschulze/devpi/commit/4dbc6ef80c2d62baf82e5e0991060936785d845d

You have the additional capability to use an installed tox as a module, do you see a large benefit for that?

wimglenn:
Well what I saw was that devpi-client was using `tox.cmdline`, so didn't technically need a tox exe to be on PATH at all.

fschulze:
The upcoming tox 4 will most likely break that, but keep the cli.

fschulze:
The upcoming devpi-client 6.0.0 will use tox as a subprocess.

PR #900, opened 2022-06-20
Enh add snapshot feature

gst:


gst:
@fschulze wdyt of this ?

.gitignore
```diff
@@ -71,3 +71,4 @@ ENV/
 /tmp/
 *.sw[po]
 **/tags
+/.idea/
```
fschulze, line 74: You can add local git ignores to ``$GIT_DIR/info/exclude``, see https://git-scm.com/docs/gitignore

server/devpi_server/keyfs_sqlite_fs.py
```diff
@@ -36,10 +36,18 @@ def __init__(self, path):
...
         self.tmppath = f"{path}-{digest}-tmp"
 
+    def __repr__(self):
+        return f"{self.__class__.__name__}({self.path})"
+
     @classmethod
     def from_content(cls, path, content_or_file):
         self = DirtyFile(path)
-        if hasattr(content_or_file, "devpi_srcpath"):
+        if hasattr(content_or_file, "filestore"):
```
fschulze, line 45: Instead of passing in a link, you should create a new link and pass in the file with the ``devpi_srcpath`` attribute set, that way it works with all storage backends like ``devpi-postgresql`` which doesn't have files on the filesystem and currently doesn't support hard links (but it could with a schema change).

server/devpi_server/main.py
```diff
@@ -462,6 +462,10 @@ def create_app(self):
         pyramid_config.add_view_deriver(self.view_deriver, under=INGRESS)
         pyramid_config.add_view_predicate('content_type', ContentTypePredicate)
 
+        # snapshoting
+        pyramid_config.add_route("/+snapshot/{user}/{index}", "/+snapshot/{user}/{index}")
```
fschulze, line 466: I would prefer the ``COPY`` HTTP method, if you have good arguments against that, then I would like ``/{user}/{index}/+copy``. Snapshotting should be an option selected with ``?snapshot``.

server/devpi_server/model.py
```diff
@@ -1511,7 +1517,11 @@ def create_linked_entry(self, rel, basename, content_or_file, last_modified=None
             assert overwrite is None
             overwrite = sum(x.get('count', 0)
                             for x in link.get_logs() if x.get('what') == 'overwrite')
-        self.remove_links(rel=rel, basename=basename)
+        if not isinstance(content_or_file, ELink):
```
fschulze, line 1520: If you create a new link and pass in a file as described in another comment, then this issue should go away.

The link removal is there to remove an overwritten file with same version if I'm not mistaken. I still need to test that.

server/devpi_server/views.py
```diff
@@ -1446,6 +1446,104 @@ def user_list(self):
...
         apireturn(200, type="list:userconfig", result=d)
 
+    @view_config(
+        route_name="/+snapshot/{user}/{index}", request_method="POST")
+    @view_config(
+        route_name="/+snapshot/{user}/{index}/", request_method="POST")
+    def snapshot(self):
+        request = self.request
+        context = self.context
+        if not request.has_permission("upload"):  # TODO: add 'snapshot' perm ?
```
fschulze, line 1456: I would use ``index_create`` for now.

Separate permissions might make sense, but then there should be ``index_copy`` enabled by default and ``index_snapshot`` disabled by default, as a admin might not want hard links to be used.

server/devpi_server/views.py
```diff
@@ -1446,6 +1446,104 @@ def user_list(self):
...
+        failures = []
+        notices = []
+        tot_wheels = 0
+        tot_project_versions = 0
+        if context.user.getstage(target_index):
+            apireturn(409, "index %r already exists" % target_index)
+        try:
+            target_stage = context.user.create_stage(target_index)
+            for project_name in projects:
+                print(f"doing {project_name} ..")
```
fschulze, line 1477: Use logging with regular string formatting and arguments instead of ``%`` operator to avoid creating the string if the log record is discarded by filters.

server/devpi_server/views.py
```diff
@@ -1446,6 +1446,104 @@ def user_list(self):
...
+                            entrypath = elink['entrypath']
+                            assert entrypath.startswith(front), elink
+                            new_elink = dict(elink)
+                            new_elink['entrypath'] = entrypath.replace(front, f"{context.user.name}/{target_index}/+f/")
+                            new_elinks.append(new_elink)
+                        except Exception as err:
+                            msg = f"error with elink {elink}: {err}"
+                            failures.append(msg)
+                            print(msg)
+                    # have to set on _data cause `verdata` itself is ReadOnly:
```
fschulze, line 1511: Check other places where version data is modified for better patterns.

server/devpi_server/views.py
```diff
@@ -1446,6 +1446,104 @@ def user_list(self):
...
+                    verdata._data['+elinks'] = new_elinks
+                    # fixup done.
+                    # we now have to set that version data into the target stage:
+                    target_stage.set_versiondata(verdata)
+                    # then we can create the (hard-)links for each release:
+                    for link in release_links:
+                        print(f"  {link.entrypath}")
+                        new_link = target_stage.store_releasefile(
+                            project_name, version,
+                            link.basename, link)
```
fschulze, line 1521: I previously wrote to create a new link. No need for that. Use ``with`` on ``file_open_read``, set ``devpi_srcpath`` when a snapshot instead of a copy is requested and pass that in. That should also keep resource usage low and the existing code for all cases is used.

fschulze (changes requested):

fschulze:
As a summary:
- I wanted to add ``copy`` anyway
- ``snapshot`` should be an option, on the server side with a query string, on the client side there should be ``copy`` and ``snapshot``.
- Try to use the pointers I gave to reuse existing functionality

I'm undecided whether we should allow this for mirror indexes. I currently lean towards forbidding it.

server/devpi_server/views.py
gst: ok. read all your comments. I'm not sure I get everything with `file_open_read` .. 

but I can see to create a new ELink in this main loop and pass that instead of the original link here .. and add a devpi_srcpath attribute to it (it does not have right now).

looking all that.. 

but FWIW: I had to `target_stage.set_versiondata(verdata)` before .. otherwise there was missing registration error.

fschulze: Instead of creating ``new_release_link`` and passing that in do something like this:

```
with link.entry.file_open_read() as f:
    f.devpi_srcpath = ...
    new_link = linkstore.create_linked_entry(
        rel="releasefile",
        basename=link.basename,
        content_or_file=f,
        last_modified=None)
    ...
```

server/devpi_server/views.py
gst: effectively `index_create` seems even more legit than upload. 

for an admin not wanting to use hard links by default.. I'm not sure to understand.. why would he prefer to have duplicate/copy of the wheels/release files instead of doing hard-links ?

server/devpi_server/views.py
gst: got it. made it like that. even better effectively. had also only to add the hash_spec and pass it. so to not have to re-compute it effectively.

fschulze:
FYI, I rebased and squashed your branch at https://github.com/fschulze/devpi/tree/enh-add-snapshot-feature. I cherry picked some changes like the ``__repr__`` stuff with some minor changes into ``main``. I still didn't have time to take a deeper look.

PR #902, opened 2022-06-23
Add favicon for the web interface

pR0Ps:
The favicon was generated from the current logo in `/devpi/doc/_static/logo.svg`. It includes 48\*48px, 32\*32px, and 16\*16px versions.

Having a favicon specified in the HTML stops browsers from trying to automatically load `/favicon.ico` which in turn attempts to look up the `favicon.ico` user, resulting in the following repeated logspam:
```
2022-06-22 11:17:08,237 ERROR [req17] [Rtx1] while handling http://127.0.0.1:3141/favicon.ico:
no user 'favicon.ico'
```

I can change this PR to use a more modern approach (PNGs) if that's preferred, I went with the `favicon.ico` just for simplicity and compatibility.


fschulze:
Does this prevent all requests for ``favicon.ico``? I've seen instances where browsers fetch it before the website itself.

Maybe we should set ``favicon.ico`` on the username blocklist and add a route for it?

pR0Ps:
In my experience, most (all?) browsers will now load the page before requesting `/favicon.ico` because the newer style of providing a favicon is dependent on parsing the path to the correct favicon out of the html. In my testing with modern Firefox, Chrome, and Safari I never saw them try to hit `/favicon.ico` after this code change while browsing the web interface. However, the browser will still try to load `/favicon.ico` when browsing a `/+simple/*` page since there's no other link for it specified on those pages.

You are correct though in that if a client wants to get the favicon without loading the page then it could still try to just load `/favicon.ico` - it's the fallback for when there is no other information so it's expected that it could still be used in some cases. This PR just fixes the majority of these cases for well-behaving browsers.

Also adding `favicon.ico` to the username blocklist and also redirecting it to the correct icon if `devpi_web` is installed sounds like a good idea.

PR #905, opened 2022-06-24
Make aiohttp use the same CA certificates as requests

pR0Ps:
By default `requests` uses CA certificates provided by the `certifi` package in order to verify HTTPS connections while `aiohttp` uses the CA certificates provided by the OS.

This change causes the `aiohttp` library to use a custom `SSLContext` that picks the CA certificates to use in the same way that `requests` does.

This fixes issues where upstream servers are trusted by one library but not another. It also allows for customizing the CA certificates to use by setting a supported environment variable.

### DIFF
--- a/server/devpi_server/main.py
+++ b/server/devpi_server/main.py
@@ -7,13 +7,16 @@ from __future__ import unicode_literals
 import aiohttp
 import inspect
 import os
+import os.path
 import asyncio
 import py
+import ssl
 import sys
 import time
 import traceback
 
 from requests import Response, exceptions
+from requests.utils import DEFAULT_CA_BUNDLE_PATH
 from devpi_common.types import cached_property
 from devpi_common.request import new_requests_session
 from .config import parseoptions, get_pluginmanager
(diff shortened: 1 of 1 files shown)


fschulze:
As far as I know the consensus for Python is to move away from custom certificate stores and use the OS one by default. I agree that there shouldn't be a difference between different libraries, but I'd rather change the store requests uses (or actually move to urllib3).

pR0Ps:
I can definitely make `requests` use a custom `HTTPAdapter` that uses an `SSLContext` created using `ssl.create_default_context()` to be exactly compatible with `aiohttp`, but I'm not sure if that's a good idea.

Changing the default from what `requests` is currently doing (`REQUESTS_CA_BUNDLE`, `CURL_CA_BUNDLE`, `certifi`) to using the system trust store will be a breaking change for at least some people. As an example, my personal devpi server started encountering issues after `aiohttp` was added because it's running on a system in which `ssl.create_default_context()` returns an `SSLContext` that doesn't verify against the correct certs.

How would you like me to proceed?

fschulze:
A well, lets cross that bridge when we get there.

PR #906, opened 2022-06-24
Avoid accessing PyPI for projects that don't exist

pR0Ps:
Previously, when a project without any cached links was requested from the `PyPIStage`, it would immediately try to load that package from the upstream server. This change introduces a check to make sure the package is actually in the stage before trying to load it.

This improves response times on slow connections, reduces the load on the upstream server, and avoids leaking potential typos of private package names to the upstream server.

### DIFF
--- a/server/devpi_server/extpypi.py
+++ b/server/devpi_server/extpypi.py
@@ -628,12 +628,14 @@ class PyPIStage(BaseStage):
                 self._offline_logging.add(project)
             return links
 
-        is_retrieval_expired = self.cache_retrieve_times.is_expired(
-            project, self.cache_expiry)
-
-        if links is None and not is_retrieval_expired:
-            raise self.UpstreamNotFoundError(
-                "cached not found for project %s" % project)
+        if links is None:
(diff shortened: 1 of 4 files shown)


fschulze:
Could you add please a changelog snippet in ``server/news``? Do you think you could add a test?

fschulze:
I just realized it would be better to use ``has_project_perstage`` and change it to use ``list_projects_perstage`` like in ``model.py``, but keep the ``is_project_cached`` check in there.

pR0Ps:
I've added a changelog snippit and moved the check into `has_project_perstage`.

Unfortunately this change seems to have broken a bunch of tests. I've added a new test that works and tests the feature, but I could use some guidance on the best way to fix the rest of them.

Right now it looks like something in tox is broken, but before I rebased on the latest main it looked like most were failing because this change implicitly requires all tests to also mock out the index to include the project instead of just mocking out the project's upstream URL.

pR0Ps:
Amazing, thanks for finishing it up for me, I was pretty unsure on how to proceed :)

Issue #913, opened 2022-07-25
[upload] workdir cleanup fails: PermissionError: Access is denied


eachimei:
devpi-client 6.0
Windows 10
Python 3.8

The upload flow (devpi.upload:main) is executed inside `with hub.workdir() as uploadbase:`:
```python
    with hub.workdir() as uploadbase:
        exported = checkout.export(uploadbase)
```
and therefore it will try to remove (`rmtree`) workdir at `__exit__` (devpi.main.Hub:workdir): 
```python
    @contextmanager
    def workdir(self, prefix='devpi-'):
        workdir = py.path.local(
            mkdtemp(prefix=prefix))

        self.info("using workdir", workdir)
        try:
            yield workdir
        finally:
            rmtree(workdir.strpath)
```

However, `checkout.export()` that runs inside this context copies also the **.git** folder, and the object files inside it are **read-only**. 
As a result, the `rmtree` command fails.

```
Traceback (most recent call last):
  File "C:\Python38\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Python38\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "<...>\.tox\release_new\lib\site-packages\devpi\__main__.py", line 3, in <module>
    main.main()
  File "<...>\.tox\release_new\lib\site-packages\devpi\main.py", line 40, in main
    return method(hub, hub.args)
  File "<...>\.tox\release_new\lib\site-packages\devpi\upload.py", line 50, in main
    uploader.do_upload_paths(archives)
  File "C:\Python38\lib\contextlib.py", line 120, in __exit__
    next(self.gen)
  File "<...>\.tox\release_new\lib\site-packages\devpi\main.py", line 239, in workdir
    rmtree(workdir.strpath)
  File "C:\Python38\lib\shutil.py", line 740, in rmtree
    return _rmtree_unsafe(path, onerror)
  File "C:\Python38\lib\shutil.py", line 613, in _rmtree_unsafe
    _rmtree_unsafe(fullname, onerror)
  File "C:\Python38\lib\shutil.py", line 613, in _rmtree_unsafe
    _rmtree_unsafe(fullname, onerror)
  File "C:\Python38\lib\shutil.py", line 613, in _rmtree_unsafe
    _rmtree_unsafe(fullname, onerror)
  [Previous line repeated 1 more time]
[...]

PR #916, opened 2022-08-07
Fix for issue #913

eachimei:


client/testing/test_upload.py
```diff
@@ -111,6 +111,9 @@ def test_vcs_export(self, uploadhub, repo, setupdir, tmpdir, monkeypatch):
             assert newrepo.join(repo.basename).join(".hg").listdir()
         else:
             assert newrepo.join(repo.basename).join(".git").listdir()
+        with uploadhub.workdir() as uploadbase:
+            exported = checkout.export(uploadbase)
```
fschulze, line 115: You should make a file read-only here. The test should then fail if you don't add the ``onerror``.

fschulze (changes requested):
Could you add a news snippet in ``client/news``, see ``examples.rst`` there.

client/testing/test_upload.py
eachimei: Done

eachimei:
> Could you add a news snippet in `client/news`, see `examples.rst` there.

Done

eachimei:
@fschulze I have problems running the tests locally, so I'm looking forward to see the unit-tests results as part of the CI

fschulze:
It seems on macOS files are deleted regardless of the read only state. I'm merging as is.

eachimei:
Thank you @fschulze 😀

PR #938, opened 2022-12-12
Add version constraint <=21.3 to dependency packaging

fyhertz:
Release 22.0 of `packaging` was published on december 8 and it breaks `devpi` for many packages (`pytz` being one of them).


fschulze:
Fixed with released devpi-common 3.7.1.

PR #959, opened 2023-03-12
Replace deprecated BadZipfile with BadZipFile

hugovk:
`BadZipfile` (with a small `f`) has been deprecated since Python 3.2, use `BadZipFile` (big `F`) instead, added in 3.2.

* https://docs.python.org/3/library/zipfile.html#zipfile.BadZipfile
* https://github.com/python/cpython/issues/86437

### DIFF
--- a/common/devpi_common/archive.py
+++ b/common/devpi_common/archive.py
@@ -32,7 +32,7 @@ def Archive(path_or_file):
     try:
         try:
             return ZipArchive(f)
-        except zipfile.BadZipfile:
+        except zipfile.BadZipFile:
             f.seek(0)
             try:
                 return TarArchive(f)


fschulze:
Thanks! I wonder why it doesn't show up as a DeprecationWarning, probably because emitting deprecation warnings for module attributes is hard (or even impossible without major workarounds). Currently devpi-common still needs to support Python 2.7. I pulled the fix into my [common4 branch](https://github.com/fschulze/devpi/tree/common4), so it will be in the next major release which will drop Python 2 support.

fschulze:
Actually there is __getattr__ for modules since Python 3.7: https://peps.python.org/pep-0562/, so I wonder why they don't use that.

hugovk:
At the moment it's only deprecated via docs (https://docs.python.org/3/library/zipfile.html#zipfile.BadZipfile). We were considering adding a `DeprecationWarning` in 3.12, but on reflection will probably leave it as a docs warning only (https://github.com/python/cpython/issues/86437).

PR #966, opened 2023-04-20
client: Use build instead of pep517

s-t-e-v-e-n-k:
pep517 has been renamed to pyproject-hooks, and as a consequence all of the deprecated functionality has been removed. Users of that functionality should switch to using build. Switch to using a method under build.util to determine the wheel metadata for upload.

### DIFF
--- a/client/devpi/upload.py
+++ b/client/devpi/upload.py
@@ -5,8 +5,8 @@ import py
 import re
 import zipfile
 
+import build.util
 import check_manifest
-import pep517.meta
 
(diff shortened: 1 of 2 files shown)


fschulze:
Thanks. Unfortunately this has to wait till the next major release, as we still support Python 2.7 and 3.5 for which there is no ``build`` version with the ``util`` module.

As far as I understand this is just future proving though. The ``pep517`` package is discontinued but won't go away.

fschulze:
Released with devpi-client 7.0.0.

PR #984, opened 2023-07-14
Fix #983: plugin hook to provide auth for mirrors

### LINKED ISSUE #983: Support basic auth for mirrors via a plugin
I would like to use devpi backed by a base from a remote index that uses basic auth. While I can do this by simply including the username and password in the mirror_url, this is obviously not ideal from a security standpoint.

I noticed that in the code it is already the case that the username and password are already extracted from the url (by the `mirror_url_auth()` method) and re-injected into the request. It seems like it would therefore be very simple to change this function to call a plugin hook to supply the actual auth credentials, falling back on the current implementation if no plugin provides anything. It would then be very simple to implement plugins (as separate packages) to access credentials from environment variables, from the .netrc file, or from the `keyring` package, for example.

My proposal would be to create a single new plugin hook: `devpiserver_get_mirror_auth(url)` which would be passed only the `mirror_url` parameter of the mirror. The return value would have to contain the username and password: I would suggest as a `dict` for simplicity although could also be a `tuple` to be consistent with the existing `devpiserver_get_credentials()` if preferred. Plugins would return `None` if they could not provide credentials for a particular url.

I am happy to implement this myself if the maintainers indicate this would be an accepted feature: I think the `devpi` side of things would be pretty minimal. The only thing I am not too sure about is the best way to test the feature.

bennyrowland:
This commit adds a new plugin hook for devpi_server that is used to query for basic auth credentials for mirror indexes. All http gets (including async ones) now inject retrieved basic auth credentials into the submitted URL. If no plugin offers any credentials then the URL remains unchanged.


fschulze:
I think this approach might cause performance issues. The hook is now called for every request to the mirror and a hook call might be expensive (it could make a http call itself to some vault service).

My proposal after thinking about it a bit:
1. Add an ``auth_url`` keyword to ``XOM.async_httpget`` and ``XOM.httpget``
2. The ``MirrorStage.async_httpget`` and ``MirrorStage.httpget`` methods set ``auth_url`` to ``self.mirror_url``
3. Add a dictionary for looked up auth data on ``XOM`` to cache it and avoid calling the hook all the time
4. Add a new method which is called with ``auth_url``. It looks up auth data in the dictionary and if there is no data calls the hook and stores the result (can be empty). This  is used by ``async_httpget``, ``httpget`` and ``mirror_auth_url`` (via ``self.xom.methodname(self.mirror_url)``)
5. Catch 401 and 403 errors in ``XOM.async_httpget`` and ``XOM.httpget`` and clear the auth data cache to let the hook run again, that way external authentication data can be updated without a restart (this may need to be improved somehow in the future, as it retries all the time, but is good enough for now)

bennyrowland:
@fschulze, sorry that it has been a long time since I raised this issue, other things got in the way. But I would now like to get back to it. I see that in the  meantime the method for inserting auth into `httpget()` *et al.* has changed to use a header, so I guess that it makes sense to use the same approach here, and presumably that would give greater flexibility to plugins as well. Does something like the following make sense?

1. `Xom` gains a dictionary mapping `url` to `auth: str`
2. inside `Xom.httpget()` and `Xom.async_httpget()` if `extra_headers` does not contain an "Authorization" key then `Xom` tries to look one up from its dictionary. If one is available it is assigned verbatim to `extra_headers`, which means other forms than "Basic" (e.g. "Bearer") could also be used if plugins could generate them.
3. If no entry is available in the dictionary then make the request without auth.
4. If we get a 401 or 403 from any url then call the hook (doesn't matter whether we already had a cached value since it was obviously no good anyway) - if we get something that is different to any cached value then we cache the new value and resubmit the request. This way we can add the WWW-Authenticate response header as an argument to the hook.

Under this scheme only XOM internals need to change, `MirrorStage` would not be affected at all, and there is no change to the API because the auth values are only looked up inside `Xom`. The alternative would be to handle everything in the `MirrorStage` which already generates a `mirror_url_authorization_header` to pass to `Xom`. It could catch a 401/403 response itself and call the hook at that point to generate a replacement value for that header.

Let me know what you think of this idea and I will get to work on it.

fschulze:
No need to apologize. I'm thankful you want to take another look.

It would be great if you could make a prove of concept doing everything in the ``MirrorStage`` as you described. Don't bother with tests too much for now, that can follow when I grasp the implications better.

I'm especially interested in how the retry can be handled properly, specifically with bad credentials.

Allowing different headers is a good idea.

How would you handle conflicts in case several plugins return a value?

bennyrowland:
I will work on a proof-of-concept, but briefly, my idea for retries was that if we get a 401/403 then call the hook and see if it suggests a different value from the one it ran with last. If we get a new credential candidate then recursively call the function with the new credential (that will cover the initial attempt without auth, and any change to the plugins during runtime) while if the suggested credential is the same as the one previously tried then allow the 401/403 to go through to a higher level response handler as currently.

That would work ok for a single provided value but things are more challenging if we consider several plugins which could all be providing values. In that case I guess my proposed logic would look something like:

1. We have a list of plugin provided credentials which is initially None.
2. If the provided credential results in a 401/403, check for the list of credentials. If it is None, call the hook to generate a list of candidates.
3. If the list of candidates is empty, make it None so that the hook will get called again next time, then propagate the 401/403.
4. Otherwise, get the next credential from the list and recursively call the get function again.
5. If we get another 401/403, go to stage 3.
6. Otherwise, propagate the response, keeping the current credential as the one to use for future requests.

There probably ought to be some kind of curation of the plugin suggested list (making sure it is not excessively long, condensing identical credentials etc.) but that is probably rare enough in practice to be a secondary consideration.

bennyrowland:
@fschulze I have a basic version implemented on my fork here: https://github.com/bennyrowland/devpi/commit/693d52d627f6484068913c22d02152b65bfc0867
It seemed easier to show you there rather than on this PR because the two implementations have so little in common.

It currently works for my use case of connecting to a pypiserver instance (using a keyring based plugin to provide the hook implementation) but I haven't thoroughly tested it for other use cases so far (hard to set up lots of different private indexes to test against). If you have other suggestions for testing I am happy to look into them. I might also try against an Azure python package index if I can get one set up without too much pain.

One thing that did strike me is that doing this on the `MirrorStage` means that the hook gets called a lot, because the stages get created so frequently. It would be possible to move all the logic of this commit into Xom instead and have it be cached there (although as a dictionary against the url rather than just the raw credentials) which would mean that in most cases it would only have to be called once. Let me know what you think.

bennyrowland:
I am closing this in favour of its replacement at #1026

PR #994, opened 2023-09-09
Ruff: Improve performance and add upper limits on code complexity

cclauss:
By setting upper limits on code complexity we can use the code review process to force those who want to increase code complexity to justify why that complexity is required.  These can also be metrics for continuous improvement.

```
tool.ruff.line-length = 250  # Default is 88
tool.ruff.mccabe.max-complexity = 37  # Default is 10
tool.ruff.pylint.max-args = 13  # Default is 5
tool.ruff.pylint.max-branches = 43  # Default is 12
tool.ruff.pylint.max-returns = 8  # Default is 6
tool.ruff.pylint.max-statements = 106  # Default is 50
```

Also, added ruff rules for `C4` and `PERF` and ran `ruff --fix .` to speed up runtime.


fschulze:
Thanks for looking into this.

Unfortunately I only completely agree with shortening the ``PL`` selection.

Also the ``PERF102`` rule looks useful, but I doubt it makes much of a difference in the places it currently warns about.

Maybe ``C402``, ``C405`` and ``C408`` when devpi-client drops <= Python 3.6, but not auto-fixing all of them, way too many unnecessary changes, as most are in non-critical tests and I prefer ``dict`` for read- and write-ability.

All ``PERF203`` are false positives, as they would change behavior.

The ``PERF401`` list comprehensions would reduce readability of the code way too much with no apparent benefit.

The ``C901`` rule, even with the increased default value is almost never useful in my opinion. I looked into these before and refactoring just to reduce function/method length makes no sense. I often already reduce function/method size if it makes sense.

I will look into picking the above rules in a way I consider useful.

fschulze:
Updated with https://github.com/fschulze/devpi/commit/ee4996678fe7c4650c51b49079140e8f51277597 in my https://github.com/fschulze/devpi/tree/common4 branch.

PR #995, opened 2023-09-09
AppVeyor: Upgrade to image: Visual Studio 2022

cclauss:
### What are the supported versions of Python vs. https://devguide.python.org/versions ?
* Do we need to keep testing legacy Python given that Python 2 died 1,347 days ago on 1/1/2020?
* Do we need to keep testing Py3{5,6,7} given that they are all end-of-life?
---
* https://www.appveyor.com/docs/lang/python
* https://www.appveyor.com/docs/windows-images-software/#python

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,3 +1,9 @@
+# https://www.appveyor.com/docs/lang/python
+# https://www.appveyor.com/docs/windows-images-software/#python
+
+image:
+  - Visual Studio 2022
+
 environment:
   matrix:
     - PROJECT: "common"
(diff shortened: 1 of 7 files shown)


fschulze:
Thanks for looking into this. Before you spend more time on this, a question: I fixed the GitHub Actions Windows tests in my https://github.com/fschulze/devpi/tree/common4 branch (which is currently held back, because current devpi-client most likely will get one last release with a fix soon), does it make sense to keep AppVeyor at all? For me GitHub Actions is much more convenient, as I see failures more easily. If AppVeyor is worth keeping, then it is fine to only test with Python 3.8 onward with it, but don't change the tox files. Older Python versions are supported due to a paid support contract.

cclauss:
Let's drop AppVeyor.

fschulze:
Dropped in my https://github.com/fschulze/devpi/tree/common4 branch.

Issue #998, opened 2023-10-02
Provide an aiohttp-less version for server


gaborbernat:
https://github.com/aio-libs/aiohttp is very slow at adding new Python version support - so allow not using it. Already raised this https://github.com/devpi/devpi/issues/918 but got closed. Because aiohttp waits until 3.12 final and even then needs a few days/weeks to get a new release devpi is acting as a gatekeeper to test early any packages that depend on devpi (such as tox). This also makes so that one cannot upgrade to 3.12 on day 1 when running locally devpi.

fschulze:
Can you suggest an alternative async http library?

gaborbernat:
https://www.python-httpx.org/async/ is great 👍

fschulze:
Doesn’t seem to have a 3.12 compatible release either yet.

gaborbernat:
Doesn't need it, it's an universal wheel `httpx-0.25.0-py3-none-any.whl` 👍

fschulze:
I'm also thinking about ditching async completely and using [``concurrent.futures``](https://docs.python.org/3/library/concurrent.futures.html) with a thread pool instead. For our use-case that might make more sense.

gaborbernat:
Either works for me 😊

gaborbernat:
@fschulze Any progress on this, aiohttp still doesn't support 3.12 and by extension devpi neither 😢

fschulze:
Seems to work with ``aiohttp==3.9.0b0`` at least: https://github.com/devpi/devpi/actions/runs/6608310136/job/17946969395

For me this has no high priority. It is likely that I will look into ``concurrent.futures`` soon, but there is no ETA at all. Most users and customers I know only update Python when the version they are using isn't supported anymore. Also devpi-web is currently blocked by the Chameleon template library being broken on Python 3.12, even though it is pure Python, but deprecations got to it.

It looks like tox tests are running on 3.12 though: https://github.com/tox-dev/tox/actions/runs/6624163180/job/18007221072, are you skipping the test that used devpi-server or marked it xfail? I don't see anything related, but devpi-server is in the list of installed packages.

gaborbernat:
devpi server is still unusable for me with:

```
Fatal error from pip prevented installation. Full pip output in file:
    /Users/bgabor8/.local/pipx/logs/cmd_2023-10-30_14.10.43_pip_errors.log

pip failed to build package:
    aiohttp

Some possibly relevant errors from pip install:
[...]

PR #1001, opened 2023-10-05
Use httpx over aiohttp

gaborbernat:
Resolves https://github.com/devpi/devpi/issues/998

### DIFF
--- a/server/devpi_server/main.py
+++ b/server/devpi_server/main.py
@@ -3,7 +3,7 @@
 a WSGI server to serve PyPI compatible indexes and a full
 recursive cache of pypi.org packages.
 """
-import aiohttp
+import httpx
 import inspect
 import os
 import os.path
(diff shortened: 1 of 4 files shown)


fschulze:
I tried the patch with some fixes in https://github.com/fschulze/devpi/tree/httpx and currently ``test_auth_mirror_url_no_hash`` and ``test_auth_mirror_url_with_hash`` fail, because httpx exposes the authentication info in its log output. We either need to block that from our side, or need to get a fix into httpx to remove at least the password from the output.

fschulze:
@zanieb thanks for the offer. Currently httpx logs the full URL including username and password of each request. It is hard to make sure this doesn't leak out onto the disk and outside services. Such leaks are often the cause for security breaks. It would be best if the username and password would be obfuscated or if it wouldn't log at all. We could configure our logging to exclude it, but if anything changes in httpx, it might come back.

zanieb:
@fschulze I took a look at the CI logs and they're pretty hard to parse :) it looks like everything fails when your secret type is logged because it's cast to a string and that throws a warning? Could you create a simpler reproduction so we can consider the best way to elide the secrets? I agree it's less than ideal to log them.

fschulze:
@zanieb The logging in httpx happens in two places:

https://github.com/encode/httpx/blob/90d71e63e017fb5535a97c3c880d1373ae8918f4/httpx/_client.py#L1013-L1020

https://github.com/encode/httpx/blob/90d71e63e017fb5535a97c3c880d1373ae8918f4/httpx/_client.py#L1729-L1736

Any request which contains a user name and password will log that. In devpi we test if the password shows up in any log output.

You even seem to have explicit tests for the logging: ``test_logging_redirect_chain`` and ``test_logging_request``
As well as documentation:
https://github.com/encode/httpx/blob/90d71e63e017fb5535a97c3c880d1373ae8918f4/docs/logging.md?plain=1#L35

fschulze:
BTW, it is best do obfuscate the username as well as the password. I have seen cases where the username is a token and the password is empty.

zanieb:
Thanks! I'm actually asking for a more concrete example of the request you're sending. I'm just guessing the username and password are part of the URL?

fschulze:
@zanieb yes ... which just lead to a thought: I could just add an ``Authorization`` header to the request instead of passing the username and password in the URL, right?

zanieb:
That seems like a much better place for it if supported by the server! I do not think headers are logged.

fschulze:
@zanieb Thanks for triggering that insight, it seems to work.

PR #1009, opened 2023-10-15
docs metadata: set `isolated` based on args (fix #1008)

### LINKED ISSUE #1008: Incorrect version given to doczip with devpi-client version 7.0.0
Given a package that uses `setuptools-scm` using `pyproject.toml` (https://github.com/pypa/setuptools_scm#pyprojecttoml-usage),
when trying to upload docs (e.g. `devpi upload --only-docs --dry-run`), the generated doczip has an incorrect version (0.0.0 which is not the true version):
`built: dist\packagename-0.0.0.doc.zip [sphinx docs] 155kb`

This appears to be due to:
https://github.com/devpi/devpi/blob/2cf7cf1d977647ba15dea24785c488371ca47305/client/devpi/upload.py#L400C44-L400C44

```
    def setup_name_and_version(self):
        try:
            metadata = build.util.project_wheel_metadata(
                str(self.rootpath), isolated=False)
```
When setting `isolated=True`, the issue is resolved and we get the correct version.

I think the root-cause is: `setuptools_scm` itself is not installed in the active virtual-env (it was installed in the isolated build frontend temporary environment when the package was originally built and installed using `build`...). This is a typical use-case in CI, for example when using `tox` to prepare a "release" environment that's going to run the `devpi` commands.

Experiments:
1. Environment as is, without changing anything -- issue is reproduced:
```
>>> import build.util
>>> metadata = build.util.project_wheel_metadata(".", isolated=False)
>>> metadata["version"]
'0.0.0'
```
2. Once changing to `isolated=True`:
```
>>> metadata = build.util.project_wheel_metadata(".", isolated=True)
>>> metadata["version"]
'2.0.12'   # <-- CORRECT version based on git tag...
```
3. Keeping `isolated=False`, but manually installing `setuptools_scm[toml]` beforehand:
```
>>> metadata = build.util.project_wheel_metadata(".", isolated=False)
>>> metadata["version"]
'2.0.12'   # <-- CORRECT version based on git tag...
```

Seems that when `isolated=False` then `build` is not installing the build-system `requires`... ? 
Ah, yes, confirmed: https://github.com/pypa/build/blob/main/src/build/util.py#L47
as can be seen, only when `isolated=True` it's going to install the build-requirements...

eachimei:
Fixes https://github.com/devpi/devpi/issues/1008

### DIFF
--- a/client/devpi/upload.py
+++ b/client/devpi/upload.py
@@ -397,7 +397,7 @@ class Exported:
     def setup_name_and_version(self):
         try:
             metadata = build.util.project_wheel_metadata(
-                str(self.rootpath), isolated=False)
+                str(self.rootpath), isolated=not self.args.no_isolation)
         except build.BuildBackendException as e:
             exc = '\n'.join(format_exception_only(
                 e.__class__, e))


fschulze:
Merged manually

Issue #1011, opened 2023-11-01
All >400 HTTP responses from login provoke a SystemExit, not using fatal, and are not always fatal errors for login


designerzim:
To reproduce:
Create a script to do important devpi tasks, e.g.:
```
# important_devpi_tasks.py
import sys
from devpi import main as devpi_client

def important_tasks(user, pw):
    devpi_client.main(
        ["devpi", "login", user, "--password", pw],
    )
    print("Other important tasks that follow.")

if __name__ == "__main__":
    print(f'command line: {" ".join(sys.argv)}\n')
    username, password = sys.argv[1:]
    important_tasks(username, password)

```

Create a new index, e.g.: `devpi index -c test/very-volatile`
Connect to the index: `devpi use test/very-volatile`
Delete the index: `devpi index --delete test/very-volatile`
Execute the script: `python important_devpi_tasks.py username password`

Expected Results:
```logged in 'username' at 'https://devpi.local.net/test/very-volatile', credentials valid for 10.00 hours
Other important tasks that follow.```

Actual Results:
```logged in 'username' at 'https://devpi.local.net/test/very-volatile', credentials valid for 10.00 hours
404 Not Found: The stage test/very-volatile could not be found.```

The script will exit unnecessarily after a successful login, preventing "Other important tasks that follow."

This happens at least on: python 3.11.5

Fresh virtual environment with only `pip install devpi` used to reproduce:
```
Package              Version
-------------------- ----------
aiohttp              3.8.6
aiosignal            1.3.1
argon2-cffi          23.1.0
argon2-cffi-bindings 21.2.0
async-timeout        4.0.3
attrs                23.1.0
beautifulsoup4       4.12.2
build                1.0.3
certifi              2023.7.22
cffi                 1.16.0
Chameleon            4.2.0
charset-normalizer   3.3.1
check-manifest       0.49
cmarkgfm             2022.10.27
colorama             0.4.6
defusedxml           0.7.1
devpi                2.2.0
devpi-client         7.0.2
devpi-common         4.0.2
devpi-server         6.9.2
devpi-web            4.2.1
docutils             0.20.1
frozenlist           1.4.0
hupper               1.12
idna                 3.4
iniconfig            2.0.0
itsdangerous         2.1.2
lazy                 1.6
multidict            6.0.4
nh3                  0.2.14
packaging            23.2
passlib              1.7.4
PasteDeploy          3.0.1
pip                  23.2.1
pkginfo              1.9.6
[...]

PR #1012, opened 2023-11-01
Handle 403, 404 gracefully without system exit

designerzim:
Reduces 403 and 404 to just warnings, and uses self.fatal for all other status codes instead of calling a raw `SystemExit(1)`

Connected to Issue #1011


fschulze:
Merged manually after some flake8 fixes.

PR #1023, opened 2023-12-20
Add `upload-time` support to links parsed from a JSON API

zanieb:
Adds support for including the `upload-time` field from PyPI in the JSON output. This allows package managers to filter files by upload time.

For example:

```
❯ curl -s -H 'Accept: application/vnd.pypi.simple.v1+json' 'http://localhost:3141/root/pypi/+simple/black' | jq .files\[0]
{
  "filename": "black-18.3a0-py3-none-any.whl",
  "url": "http://localhost:3141/root/pypi/+f/718/3263650ba3071/black-18.3a0-py3-none-any.whl",
  "hashes": {
    "sha256": "7183263650ba3071034e90b40a1ea74abccbd32cf525cef6d7914479dbe7f2fb"
  },
  "requires-python": ">=3.6",
  "upload-time": "2018-03-14T21:30:34.781336Z"
}
```

I'd appreciate some guidance on where to add test coverage.

### DIFF
--- a/server/devpi_server/importexport.py
+++ b/server/devpi_server/importexport.py
@@ -561,12 +561,14 @@ class Importer:
                 links = [(url.basename, entry.relpath)]
                 requires_python = [versions[version].get('requires_python')]
                 yanked = [versions[version].get('yanked')]
-                for key, href, require_python, is_yanked in links_with_data:
+                upload_time = [versions[version].get('upload_time')]
+                for key, href, require_python, is_yanked, upload_time_ in links_with_data:
                     links.append((key, href))
                     requires_python.append(require_python)
                     yanked.append(is_yanked)
+                    upload_time.append(upload_time_)
                 stage._save_cache_links(
-                    project, links, requires_python, yanked, serial, None)
+                    project, links, requires_python, yanked, upload_time, serial, None)
             # devpi-server-2.1 exported with md5 checksums
             if "md5" in mapping:
                 assert "hash_spec" not in mapping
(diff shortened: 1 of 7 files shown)


fschulze:
I would advice you to wait with this feature. I'm doing a complete refactor of the storage backend which will make this much easier. Currently it is very hard to get right with backward compatibility, replication and what have you.

fschulze:
Please keep this open though, it is a good use case to test my refactoring.

zanieb:
No problem thanks for giving it a look.

I needed this personally, otherwise my package manager tests skip all of the wheels :) but I can just install from my fork for now. 

Let me know if you need anything.

zanieb:
Let me know if you want me to revisit this.

pelson:
I was just contributing to [`pypi-timemachine`](https://github.com/astrofrog/pypi-timemachine), which is essentially doing the same thing as `uv`'s `exclude-newer`, and noted this PR - having something like this would be super useful for that use case. 

In https://github.com/devpi/devpi-constrained/issues/12 I proposed seeing if there is an option, and interest, to re-use parts of [simple-repository](https://github.com/astrofrog/pypi-timemachine) for handling some of the data model and logic. Since @fschulze suggests that there is a significant internal re-write, perhaps it is a good option to re-use such a library. I don't think `simple-repository` is perfect at this point, but it is a pretty good start, and I think integrating it into something like `devpi` would certainly lead to some improvement on the `simple-repository` side (which I would be most happy to do!).

PR #1026, opened 2024-02-20
Fix #983: plugin hook to provide auth for mirrors

### LINKED ISSUE #983: Support basic auth for mirrors via a plugin

bennyrowland:
This is a replacement for my original PR #984 based on suggestions from @fschulze. As their suggestions resulted in a complete rewrite, and there were also a large number of commits from the primary repo to merge in to my fork in the meantime, it seemed easier to me to branch off the latest `main` and submit a new PR.

In essence, this PR adds a new plugin hook for devpi_server called `devpiserver_get_mirror_auth()` which can be used to provide an "Authorization" header to be submitted with requests to mirrors (such as private, secured, package indexes). The hook gets called when devpi receives a 401 or 403 from an upstream index and any generated credentials are tried sequentially until one works. That credential will then be used for all future requests to the index - if it stops working, any remaining credentials are tried, then the response is propagated up the chain. Any future requests will then cause the hook to be called again, potentially obtaining fresh credentials.

For reference, I am currently using this with a very simple keyring based hook implementation that looks like this:
```
@hookimpl()
@suppress(keyring.errors.KeyringError)
def devpiserver_get_mirror_auth(mirror_url, www_authenticate_header):
    cred = keyring.get_credential(mirror_url.hostname, mirror_url.username)
    if cred:
        auth = f"{cred.username or ''}:{cred.password or ''}".encode()
        return f"Basic {b64encode(auth).decode()}"
    return None
```
This provides Basic auth to connect to a secured PyPIServer instance.

If you can give me some advice on the kinds of tests that would be appropriate in this case then I will implement those too.

### DIFF
--- a/server/devpi_server/hookspecs.py
+++ b/server/devpi_server/hookspecs.py
@@ -201,6 +201,18 @@ def devpiserver_auth_denials(request, acl, user, stage):
 
 
 @hookspec
+def devpiserver_get_mirror_auth(mirror_url, www_authenticate_header):
+    """Provide an http "Authorization" header to access a mirror.
+
+    Return a string which will be set as the authorization header for all http
+    requests to this mirror.
+
+    Return None or an empty string if no credentials can be determined for the
+    supplied url.
+    """
+
+
+@hookspec
 def devpiserver_get_stage_customizer_classes():
     """EXPERIMENTAL!
 
--- a/server/devpi_server/mirror.py
+++ b/server/devpi_server/mirror.py
@@ -250,11 +250,19 @@ class MirrorStage(BaseStage):
         # if we have any auth candidates, the first one has just failed, so
         # discard it, allowing any others in the list to be tried
         # if we didn't have any auth candidates, try and obtain some
-        if self.auth_candidates:
-            self.auth_candidates.pop(0)
+        auth_candidates = self.xom.setdefault_singleton(
+            self.name, "auth_candidates", factory=list
+        )
+        if auth_candidates:
+            auth_candidates.pop(0)
         else:
             pm = get_pluginmanager()
-            self.auth_candidates = pm.hook.devpiserver_get_mirror_auth(mirror_url=self.mirror_url, www_authenticate_header=auth_header)
+            auth_candidates.extend(
+                pm.hook.devpiserver_get_mirror_auth(
+                    mirror_url=self.mirror_url,
+                    www_authenticate_header=auth_header
+                )
+            )
 
 
     @property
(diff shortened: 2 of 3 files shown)


bennyrowland:
@fschulze ok, I may have been a little premature in thinking that this solved the problem completely. The authorisation here is getting through to requests on the MirrorStage for lists of packages and versions, but when it comes to actually serving the packages everything happens through a direct call to `xom.httpget()` in `iter_cache_remote_file()` from `_pkgserv()`. `_pkgserv()` injects basic auth into the url from its stage's `mirror_url_auth` property, but doesn't get the "Authorization" header.

This means that there are basically two options for handling this: move all the auth information from MirrorStage to Xom or replace all bare calls to `xom.[async_]httpget()` with `stage.[async_]httpget()` ones.

1. currently the data is stored on the Xom singleton, but keyed to the MirrorStage name, we could change the key to something else (presumably part of the URL but not sure exactly which part as in principle there could be two remote indexes hosted under different parts of the same URL) and retrieve it inside `Xom.[async_]httpget()`. Very easy if we can work out the correct key.
2. there are currently only 3 calls to `xom.[async_]httpget()` that I can find that don't go through MirrorStage, they are in `iter_cache_remote_file()` as described above, and twice in `iter_remote_file_replica()`, both in `views.py`. Both of these functions only seem to be called from `_pkgserv()` (via `iter_fetch_remote_file()`) so I think that they should both always have access to a `MirrorStage` which could be passed down the function chain instead of the `Xom`. This would also then let us strip out a chunk of the current basic auth extracting code from all those places as well.

Let me know which of these options you prefer (I think that option 2 will be easier but I don't know all the details, particularly about replicas). Also keen to get some general feedback on the PR overall, and what it will take to get it accepted.
Thanks.

fschulze:
I ended up adding ``test_auth_mirror_plugin`` while I was trying to provide just a base skeleton for a test. I made some fixes and merged it manually as one commit. It would be great if you could try to add some more tests with other cases like multiple plugins or no valid auth data.

PR #1034, opened 2024-04-15
Fix check for extracted path when testing packages where the package …

aalcalajitc:
…name contains a dot, but the extracted path has underscores.

Noticed this happening with latest `setuptools` release with changes for PE625, it causes `devpi test` to fail with an error saying it could not find unpacked package, this PR adds a check for this case.


fschulze:
Thanks! There were a few more cases causing issues. I parametrized the test and generalized the check in ``unpack`` using globbing. I manually merged your commit together with my changes.

PR #1045, opened 2024-07-16
Add support for UV pip

bartv:
This PR add uv pip useragent to the list of supported installer. This makes it easier to use devpi with uv pip


fschulze:
Thanks! I added uv to the relevant test parameters and moved the changelog entry. This will go into the upcoming 6.13.0.

PR #1058, opened 2024-09-06
Fixing bug in export to temp directory on folders.

### LINKED ISSUE #1057: devpi upload fails on version 7.0 and later with PermissionError on folders
On the devpi-client 7.0 and later, the upload command will fail with a permission error.

Steps to reproduce are as simple as cloning the current repo and running `devpi upload`:

image

The call to shutil.copy2 treats folder paths like '.ci' or '.github' as files and not folders.

anthony-n-brown:
shtuil.copy2 will throw PermissionError when the destination is a folder. Seeing as we always make parent directories before running copy, there's no point in processing any directories.

Fixes #1057

### DIFF
--- a/client/devpi/upload.py
+++ b/client/devpi/upload.py
@@ -350,6 +350,8 @@ class Checkout:
         for fn in files:
             source = self.rootpath / fn
             dest = newrepo / fn
+            if os.path.isdir(source):
+                continue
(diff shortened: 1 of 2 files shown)


anthony-n-brown:
Sorry for the delay, had notifications disabled :(

You adjusted change works just fine as well.

fschulze:
Merged manually on main.

PR #1070, opened 2024-11-27
Add a server option to pass a custom connection limit to waitress

julienjpk-withings:
devpi-server currently runs with a hardcoded limit of 100 simulatenous connections, enforced through its waitress dependency. This can be limiting in production environments where devpi may need to handle more, or where connections are kept open by a pooler (load balancer, front HTTP server, ...).

This PR adds a `--connection-limit` parameter allowing administrators to tailor this limit to their needs and resources.


fschulze:
Thanks!

PR #1083, opened 2025-03-05
Add PostgreSQL custom schema support

bcvswiss:
# Add PostgreSQL Schema Support

Added support for custom PostgreSQL schemas in devpi-postgresql. Users can now specify a schema other than `public` via the connection string.

## Changes

- Added `schema` parameter to `Storage` class (default: `public`)
- Modified code to check for and create tables in the specified schema
- Updated documentation

## Testing

Tested with custom schemas, default schema, and server restart scenarios. All tests pass.

## Checklist

- [x] Code follows project style
- [x] Tests added and passing
- [x] Documentation updated
- [x] Backward compatibility maintained

### DIFF
--- a/postgresql/README.rst
+++ b/postgresql/README.rst
@@ -37,9 +37,24 @@ Upon first initialization of ``devpi-server`` use ``--storage pg8000`` to select
 
 By default it'll use the ``devpi`` database on ``localhost`` port ``5432``.
 To change that, use ``storage pg8000:host=example.com,port=5433,database=devpi_prod``.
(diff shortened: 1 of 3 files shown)


bcvswiss:
Let me know if this feature is needed, or else we can close the PR @fschulze.

fschulze:
I think the feature is interesting.
On the implementation side I would split ``ensure_tables_exist`` to have the try/except block separate and remove the added indentation. The added indentation also makes it hard to read the diff of that function in the commit, so it would be nice if you can rewrite the commit if you do so.
I'm also not sure yet what the changes in ``get_connection`` are for. Could be that the intent is also a bit obscured by added indentation for try/expect?

PR #1120, opened 2026-04-13
fix: cache uv get package request fail

### LINKED ISSUE #1119: cache some uv get package request failed
I am using devpi with uv, sometimes uv download other platform whl file (for lock), dont know why not all of them, for some package uv will not complete the request(might just lack one or two last packet),the cache process will terminal at 99%, and I still have to download whl again when install in another project.

this is a part of log.

```
devpi:
2026-04-13 19:43:02,504 DEBUG [req111] 200 OK 0.014s serial=25 length=None type=application/vnd.pypi.simple.v1+json
2026-04-13 19:43:02,507 INFO  Client disconnected while serving /root/pypi/+f/db0/dbfd2a6cdf377/pandas-3.0.2-cp314-cp314-macosx_10_15_x86_64.whl
2026-04-13 19:43:02,509 INFO  [req112] GET /root/pypi/+f/4b1/d2be7ac37ceaf/tzdata-2026.1-py2.py3-none-any.whl
2026-04-13 19:43:02,510 DEBUG [req112] [Rtx25] in-transaction 25
2026-04-13 19:43:02,513 DEBUG [req112] [Rtx25] closing transaction at 25
2026-04-13 19:43:02,514 DEBUG [req112] 200 OK 0.004s serial=25 length=348952 type=None

uv:
DEBUG Skipping file for pandas: pandas-0.9.1.win32-py2.5.exe
DEBUG Skipping file for pandas: pandas-0.9.1.win32-py2.6.exe
DEBUG Skipping file for pandas: pandas-0.9.1.win32-py2.7.exe
DEBUG Skipping file for pandas: pandas-0.9.1.win32-py3.1.exe
DEBUG Skipping file for pandas: pandas-0.9.1.win32-py3.2.exe
DEBUG No cache entry for: http://localhost:3141/root/pypi/+f/db0/dbfd2a6cdf377/pandas-3.0.2-cp314-cp314-macosx_10_15_x86_64.whl
DEBUG Sending fresh HEAD request for: http://localhost:3141/root/pypi/+f/db0/dbfd2a6cdf377/pandas-3.0.2-cp314-cp314-macosx_10_15_x86_64.whl
```

YAGregor:
continue caching package when client disconnect request. fix: #1119

### DIFF
--- a/server/devpi_server/views.py
+++ b/server/devpi_server/views.py
@@ -1804,6 +1804,12 @@ def _headers_from_response(r):
 
 
 class FileStreamer:
+    _data_iter: Iterator[bytes] | None
+    _running_hashes: RunningHashes | None
+    _file_size: int
+    _content_size: int | None
+    _download_completed: bool
+
     def __init__(self, f, entry, response):
         self.hash_type = entry.best_available_hash_type
         self.hash_types = entry.default_hash_types
@@ -1812,32 +1818,58 @@ class FileStreamer:
         self.response = response
         self.error = None
         self.f = f
+        self._data_iter = None
+        self._running_hashes = None
+        self._file_size = 0
+        self._content_size = None
+        self._download_completed = False
 
     def __iter__(self):
-        filesize = 0
+        self._file_size = 0
         running_hashes = RunningHashes(self.hash_type, *self.hash_types)
+        self._running_hashes = running_hashes
         running_hashes.start()
-        content_size = self.response.headers.get("content-length")
+        self._content_size = self.response.headers.get("content-length")
 
         yield _headers_from_response(self.response)
 
         data_iter = self.response.iter_raw(10240)
+        self._data_iter = data_iter
         while 1:
             data = next(data_iter, None)
             if data is None:
                 break
-            filesize += len(data)
+            self._file_size += len(data)
             for rh in running_hashes._running_hashes:
                 rh.update(data)
             self.f.write(data)
             yield data
 
-        self.hashes = running_hashes.digests
+        self.save_file_and_gen_hash()
 
-        if content_size and int(content_size) != filesize:
+    def save_file_and_gen_hash(self):
+        if self._download_completed:
+            return
+
+        running_hashes = self._running_hashes
+        data_iter = self._data_iter
+        while 1 and (data_iter is not None):
+            data = next(data_iter, None)
+            if data is None:
+                break
+            self._file_size += len(data)
+            for rh in running_hashes._running_hashes:
+                rh.update(data)
+            self.f.write(data)
+
+        self.hashes = running_hashes.digests
+        content_size = self._content_size
+        self._download_completed = True
+
+        if content_size and int(content_size) != self._file_size:
             raise ValueError(
                 "%s: got %s bytes of %r from remote, expected %s" % (
-                    self.relpath, filesize, self.response.url, content_size))
+                    self.relpath, self._file_size, self.response.url, content_size))
         if self._hashes:
             err = self.hashes.exception_for(self._hashes, self.relpath)
             if err is not None:
@@ -1867,8 +1899,7 @@ def iter_cache_remote_file(stage, entry, url):
             raise
         except GeneratorExit:
             threadlog.error("client disconnected, still continue update cache")
-            for _ in file_streamer:
-                pass
+            file_streamer.save_file_and_gen_hash()
 
         if not entry.has_existing_metadata():
             with xom.keyfs.write_transaction(allow_restart=True):
--- a/server/devpi_server/views.py
+++ b/server/devpi_server/views.py
@@ -1865,6 +1865,10 @@ def iter_cache_remote_file(stage, entry, url):
         except Exception as err:
             threadlog.error(str(err))
             raise
+        except GeneratorExit:
+            threadlog.error("client disconnected, still continue update cache")
+            for _ in file_streamer:
+                pass
 
         if not entry.has_existing_metadata():
             with xom.keyfs.write_transaction(allow_restart=True):
--- a/server/test_devpi_server/test_streaming.py
+++ b/server/test_devpi_server/test_streaming.py
@@ -192,3 +192,4 @@ class TestStreaming(object):
                 break
             sleep(0.1)
         assert pkg_file.exists()
+        assert pkg_file.read_bytes() == content
--- a/server/test_devpi_server/test_streaming.py
+++ b/server/test_devpi_server/test_streaming.py
@@ -1,8 +1,17 @@
+from __future__ import annotations
+
+from time import sleep
 import base64
 import contextlib
 import httpx
 import pytest
 import sys
+import typing
+
+
+if typing.TYPE_CHECKING:
+    from testing.simpypi import SimPyPI
+    import pathlib
 
 
 pytestmark = [
@@ -58,17 +67,9 @@ class TestStreaming(object):
     @pytest.mark.slow
     @pytest.mark.parametrize("length,pkg_version,pkg_name", [
         (None, '1.0', 'pkg1'), (False, '1.1', 'pkg2')])
-    def test_streaming_download(
-        self,
-        content_digest,
-        files_path,
-        length,
-        pkg_version,
-        pkg_name,
-        server_url_session,
-        simpypi,
-    ):
-        from time import sleep
+    def test_streaming_download(self, content_digest, files_path, length, pkg_version, pkg_name, server_url_session, simpypi, storage_info):
+        if "storage_with_filesystem" not in storage_info.get('_test_markers', []):
+            pytest.skip("The storage doesn't have marker 'storage_with_filesystem'.")
         (content, digest) = content_digest
         (url, s) = server_url_session
         pkgzip = f"{pkg_name}-{pkg_version}.zip"
@@ -139,3 +140,55 @@ class TestStreaming(object):
         pkg_file = files_path.joinpath(
             'root', 'pypi', '+f', digest[:3], digest[3:16], pkgzip)
         assert not pkg_file.exists()
+
+    @pytest.mark.slow
+    @pytest.mark.parametrize(
+        "length,pkg_version,pkg_name,disconnect_at_length", [(None, "1.0", "pkg1", 10), (False, "1.1", "pkg2", -1025)]
+    )
+    def test_disconnect_while_streaming(
+        self,
+        content_digest: tuple[bytes, bytes],
+        files_path: pathlib.Path,
+        length: bool | None,  # noqa: FBT001
+        pkg_version: str,
+        pkg_name: str,
+        server_url_session: tuple[str, httpx.Client],
+        simpypi: SimPyPI,
+        storage_info: dict,
+        disconnect_at_length: int
+    ):
+        from time import sleep
+
+        if "storage_with_filesystem" not in storage_info.get("_test_markers", []):
+            pytest.skip("The storage doesn't have marker 'storage_with_filesystem'.")
+        (content, digest) = content_digest
+        (url, s) = server_url_session
+        pkgzip = f"{pkg_name}-{pkg_version}.zip"
+        simpypi.add_release(pkg_name, pkgver="%s#sha256=%s" % (pkgzip, digest))
+        simpypi.add_file(f"/{pkg_name}/{pkgzip}", content, stream=True, length=length)
+        with contextlib.closing(s.get(url + f"root/mirror/{pkg_name}")) as r:
+            r = r.json()
+        assert pkg_version in r["result"], r
+        href = r["result"][pkg_version]["+links"][0]["href"]
+        data = b''
+        receive_bytes_len = 0
+        disconnect_len = disconnect_at_length if disconnect_at_length > 0 else len(content) + disconnect_at_length
+        with httpx.stream("get", href) as r:
+            stream = r.iter_bytes(1024)
+            while True:
+                streaming_data = next(stream)
+                data += streaming_data
+                if len(data) >= disconnect_len:
+                    break
+
+        assert data != content
+
+        pkg_file = files_path.joinpath(
+            "root", "mirror", "+f", digest[:3], digest[3:16], pkgzip
+        )
+        # this is sometimes delayed a bit, so we check for a while
+        for i in range(50):
+            if pkg_file.exists():
+                break
+            sleep(0.1)
+        assert pkg_file.exists()


fschulze:
This fix doesn't seem to work. I tried to write a test for it and there are always other errors later. First issue is that the ``for _ in file_streamer:`` loop restarts the iteration. But even storing the iterator with ``iter(file_streamer)`` and yielding from that and looping over the remains after ``GeneratorExit`` is still interrupted and the file is never fully written.

```diff
diff --git a/server/test_devpi_server/simpypi.py b/server/test_devpi_server/simpypi.py
index 300b4fa877..4e1de3f239 100644
--- a/server/test_devpi_server/simpypi.py
+++ b/server/test_devpi_server/simpypi.py
@@ -96,16 +96,20 @@ class SimPyPIRequestHandler(httpserver.BaseHTTPRequestHandler):
             else:
                 chunksize = f['chunksize']
                 callback = f.get('callback')
-                for i in range(len(content) // chunksize):
-                    data = content[i * chunksize:(i + 1) * chunksize]
-                    if not data:
-                        break
-                    self.wfile.write(data)
-                    if callback:
-                        callback(i * chunksize)
-                    simpypi.add_log(
-                        "do_GET", self.path,
-                        "streamed %i bytes" % (i * chunksize))
+                try:
+                    self.server.is_streaming = True
+                    for i in range(len(content) // chunksize):
+                        data = content[i * chunksize:(i + 1) * chunksize]
+                        if not data:
+                            break
+                        self.wfile.write(data)
+                        if callback:
+                            callback(i * chunksize)
+                        simpypi.add_log(
+                            "do_GET", self.path,
+                            "streamed %i bytes" % (i * chunksize))
+                finally:
+                    self.server.is_streaming = False
                 return
         simpypi.add_log("do_GET", self.path, "not found")
         start_response(404, headers)
@@ -118,6 +122,7 @@ class SimPyPI:
         self.projects = {}
         self.files = {}
         self.clear()
+        self.is_streaming = False
 
     def clear(self):
         self.clear_log()
diff --git a/server/test_devpi_server/test_streaming.py b/server/test_devpi_server/test_streaming.py
index 73c2a93878..4acc91c575 100644
--- a/server/test_devpi_server/test_streaming.py
+++ b/server/test_devpi_server/test_streaming.py
@@ -1,3 +1,4 @@
+from time import sleep
 import base64
 import contextlib
 import httpx
@@ -68,7 +69,6 @@ class TestStreaming(object):
         server_url_session,
         simpypi,
     ):
-        from time import sleep
         (content, digest) = content_digest
         (url, s) = server_url_session
         pkgzip = f"{pkg_name}-{pkg_version}.zip"
@@ -139,3 +139,42 @@ class TestStreaming(object):
         pkg_file = files_path.joinpath(
             'root', 'pypi', '+f', digest[:3], digest[3:16], pkgzip)
         assert not pkg_file.exists()
+
+    @pytest.mark.parametrize("size_factor,pkg_version,pkg_name", [
+        (2, '1.2', 'pkg3'), (0.5, '1.3', 'pkg4')])
+    def test_streaming_download_aborted(
+        self,
+        content_digest,
+        files_path,
+        pkg_version,
+        pkg_name,
+        server_url_session,
+        simpypi,
+        size_factor,
+    ):
+        (content, digest) = content_digest
+        (url, s) = server_url_session
+        pkgzip = f"{pkg_name}-{pkg_version}.zip"
+        length = int(len(content) * size_factor)
+        simpypi.add_release(pkg_name, pkgver='%s#sha256=%s' % (pkgzip, digest))
+        simpypi.add_file(
+            f"/{pkg_name}/{pkgzip}", content, stream=True, length=length)
+        with contextlib.closing(s.get(url + f"root/mirror/{pkg_name}")) as _r:
+            r = _r.json()
+        assert pkg_version in r['result'], r
+        href = r['result'][pkg_version]['+links'][0]['href']
+        simpypi.clear_requests()
+        with httpx.stream("get", href) as r:
+            stream = r.iter_bytes(1024)
+            part = next(stream)
+            assert part == b'deadbeaf' * 128
+            part = next(stream)
+            assert part == b'sandwich' * 128
+            # now we just stop short
+        while 1:
+            if not simpypi.is_streaming:
+                break
+            sleep(0.1)
+        pkg_file = files_path.joinpath(
+            'root', 'pypi', '+f', digest[:3], digest[3:16], pkgzip)
+        assert pkg_file.exists()
```

fschulze:
I doubt you ran the tests, because this is just what I did, which didn't fix the problem.

YAGregor:
I am considering a fix.I make this on windows which seems tests can't be ran on.I may split FileStreamer into two part, real time streaming and file download, some variable may change to property (with type hint) to continue download file.

fschulze:
The point of the streaming is that the download and the serving to the client happens in parallel, if you do it one after the other, the client will likely time out. Keep that in mind when doing the changes.

The tests can kinda be run on Windows if you use the WSL (Windows Subsystem for Linux) stuff. Some tests require nginx, but they should be skipped automatically if it isn't available.

YAGregor:
> This fix doesn't seem to work. I tried to write a test for it and there are always other errors later. First issue is that the `for _ in file_streamer:` loop restarts the iteration. But even storing the iterator with `iter(file_streamer)` and yielding from that and looping over the remains after `GeneratorExit` is still interrupted and the file is never fully written.
> 
> ```diff
> diff --git a/server/test_devpi_server/simpypi.py b/server/test_devpi_server/simpypi.py
> index 300b4fa877..4e1de3f239 100644
> --- a/server/test_devpi_server/simpypi.py
> +++ b/server/test_devpi_server/simpypi.py
> @@ -96,16 +96,20 @@ class SimPyPIRequestHandler(httpserver.BaseHTTPRequestHandler):
>              else:
>                  chunksize = f['chunksize']
>                  callback = f.get('callback')
> -                for i in range(len(content) // chunksize):
> -                    data = content[i * chunksize:(i + 1) * chunksize]
> -                    if not data:
> -                        break
> -                    self.wfile.write(data)
> -                    if callback:
> -                        callback(i * chunksize)
> -                    simpypi.add_log(
> -                        "do_GET", self.path,
> -                        "streamed %i bytes" % (i * chunksize))
> +                try:
> +                    self.server.is_streaming = True
> +                    for i in range(len(content) // chunksize):
> +                        data = content[i * chunksize:(i + 1) * chunksize]
> +                        if not data:
> +                            break
> +                        self.wfile.write(data)
> +                        if callback:
> +                            callback(i * chunksize)
> +                        simpypi.add_log(
> +                            "do_GET", self.path,
> +                            "streamed %i bytes" % (i * chunksize))
> +                finally:
> +                    self.server.is_streaming = False
>                  return
>          simpypi.add_log("do_GET", self.path, "not found")
>          start_response(404, headers)
> @@ -118,6 +122,7 @@ class SimPyPI:
>          self.projects = {}
>          self.files = {}
>          self.clear()
> +        self.is_streaming = False
>  
>      def clear(self):
>          self.clear_log()
> diff --git a/server/test_devpi_server/test_streaming.py b/server/test_devpi_server/test_streaming.py
> index 73c2a93878..4acc91c575 100644
> --- a/server/test_devpi_server/test_streaming.py
> +++ b/server/test_devpi_server/test_streaming.py
> @@ -1,3 +1,4 @@
> +from time import sleep
>  import base64
>  import contextlib
>  import httpx
> @@ -68,7 +69,6 @@ class TestStreaming(object):
>          server_url_session,
>          simpypi,
>      ):
> -        from time import sleep
>          (content, digest) = content_digest
>          (url, s) = server_url_session
>          pkgzip = f"{pkg_name}-{pkg_version}.zip"
> @@ -139,3 +139,42 @@ class TestStreaming(object):
>          pkg_file = files_path.joinpath(
>              'root', 'pypi', '+f', digest[:3], digest[3:16], pkgzip)
>          assert not pkg_file.exists()
> +
> +    @pytest.mark.parametrize("size_factor,pkg_version,pkg_name", [
> +        (2, '1.2', 'pkg3'), (0.5, '1.3', 'pkg4')])
> +    def test_streaming_download_aborted(
> +        self,
> +        content_digest,
> +        files_path,
> +        pkg_version,
> +        pkg_name,
> +        server_url_session,
> +        simpypi,
> +        size_factor,
> +    ):
> +        (content, digest) = content_digest
> +        (url, s) = server_url_session
> +        pkgzip = f"{pkg_name}-{pkg_version}.zip"
> +        length = int(len(content) * size_factor)
> +        simpypi.add_release(pkg_name, pkgver='%s#sha256=%s' % (pkgzip, digest))
> +        simpypi.add_file(
> +            f"/{pkg_name}/{pkgzip}", content, stream=True, length=length)
> +        with contextlib.closing(s.get(url + f"root/mirror/{pkg_name}")) as _r:
> +            r = _r.json()
> +        assert pkg_version in r['result'], r
> +        href = r['result'][pkg_version]['+links'][0]['href']
> +        simpypi.clear_requests()
> +        with httpx.stream("get", href) as r:
> +            stream = r.iter_bytes(1024)
> +            part = next(stream)
> +            assert part == b'deadbeaf' * 128
> +            part = next(stream)
> +            assert part == b'sandwich' * 128
> +            # now we just stop short
> +        while 1:
> +            if not simpypi.is_streaming:
> +                break
> +            sleep(0.1)
> +        pkg_file = files_path.joinpath(
> +            'root', 'pypi', '+f', digest[:3], digest[3:16], pkgzip)
> +        assert pkg_file.exists()
> ```

carefully read this test, I do not think this test case is correct because you copied the `test_streaming_diff_size`, in this case the response can never be download complete (because response real length != Content-Length). I have written a new test and let cache refreshing process continue when request close during client download.

fschulze:
Good catch on my copy/paste error in the test.

After that I took another look and read up on GeneratorExit. I found https://stackoverflow.com/questions/30862196/generatorexit-in-python-generator which lead me to a simpler solution solely in FileStreamer without repeating the loop. That also automatically fixes this for fetching in replicas from primary.

https://github.com/fschulze/devpi/commit/802307b13cd7af93c30338050157ba2320953282

fschulze:
Manually merged. Will be in next release.

YAGregor:
I think we may missing something @fschulze  , GeneratorExit should not be catch. otherwise it may cause runtime error, as the python document 

> catching the exception and doing anything else is illegal and will trigger a [RuntimeError](https://docs.python.org/3.10/library/exceptions.html#RuntimeError). [close()](https://docs.python.org/3.10/reference/expressions.html#generator.close) will also be called by Python’s garbage collector when the generator is garbage-collected.

> If you need to run cleanup code when a [GeneratorExit](https://docs.python.org/3.10/library/exceptions.html#GeneratorExit) occurs, I suggest using a try: ... finally: suite instead of catching [GeneratorExit](https://docs.python.org/3.10/library/exceptions.html#GeneratorExit).

https://docs.python.org/3.10/howto/functional.html#passing-values-into-a-generator

fschulze:
Then your fix wouldn't be ok either. As far as I understand you only get RuntimeError when you continue to yield. When we are actually garbage collected, then there will be other errors because of missing references etc. I will try it for a while before a release.

YAGregor:
> Then your fix wouldn't be ok either. As far as I understand you only get RuntimeError when you continue to yield. When we are actually garbage collected, then there will be other errors because of missing references etc. I will try it for a while before a release.

Yes @fschulze  , I'd like to write a new version. Would you prefer to handle this on your own, or wait for my upcoming version?

fschulze:
If you come up with something better, go ahead. I think the current solution is ok. I really wonder how you will come up with something which doesn't catch ``GeneratorExit``. I think it would be possible to supply a custom class with ``__next__`` and ``close``, but it would have the same issue in differentating between the ``close`` call from ``waitress`` and a ``close`` call from the garbage collector.

fschulze:
Another option may be a wrapping iterator which will be closed by waitress, but the inner iterator can continue.

YAGregor:
I've written a new version https://github.com/YAGregor/devpi/commit/4bbe5f33f8f337b6f6a3e7b72397c74e6d4127e0 without catch GeneratorExit and copy paste, the key idea is 
* treat continue download as "clean up" in finally
* object is poorman's closure, move the variables to property to keep stream state
* flatten code struct by helper function

tried your version https://github.com/fschulze/devpi/commit/802307b13cd7af93c30338050157ba2320953282 , it can't pass new test case 

```
test_devpi_server/test_streaming.py:151 (TestStreaming.test_disconnect_while_streaming[None-1.0-pkg1-10])
self = <test_devpi_server.test_streaming.TestStreaming object at 0x72f2cf9081f0>
content_digest = (b'deadbeafdeadbeafdeadbeafdeadbeafdeadbeafdeadbeafdeadbeafdeadbeafdeadbeafdeadbeafdeadbeafdeadbeafdeadbeafdeadbeafdea...hsandwichsandwichsandwichsandwichsandwichsandwich', '77d640a11ff574a8bd5bcd6c2da34b7f0cef75a4974856fdae734a03b134b70d')
files_path = PosixPath('/tmp/test-szj_5id3-server-directory/primary/+files')
length = None, pkg_version = '1.0', pkg_name = 'pkg1'
server_url_session = ('http://localhost:45395/', <httpx.Client object at 0x72f2cf78a5c0>)
simpypi = <test_devpi_server.simpypi.SimPyPI object at 0x72f2cf78a2f0>
disconnect_at_length = 10

    @pytest.mark.slow
    @pytest.mark.parametrize(
        ("length", "pkg_version", "pkg_name", "disconnect_at_length"),
        [(None, "1.0", "pkg1", 10), (False, "1.1", "pkg2", -1025)],
    )
    def test_disconnect_while_streaming(
        self,
        content_digest: tuple[bytes, str],
        files_path: pathlib.Path,
        length: bool | None,  # noqa: FBT001
        pkg_version: str,
        pkg_name: str,
        server_url_session: tuple[str, httpx.Client],
        simpypi: SimPyPI,
        disconnect_at_length: int,
    ) -> None:
        (content, digest) = content_digest
        (url, s) = server_url_session
        pkgzip = f"{pkg_name}-{pkg_version}.zip"
        simpypi.add_release(pkg_name, pkgver="%s#sha256=%s" % (pkgzip, digest))
        simpypi.add_file(f"/{pkg_name}/{pkgzip}", content, stream=True, length=length)
        with contextlib.closing(s.get(url + f"root/mirror/{pkg_name}")) as _r:
            r = _r.json()
        assert pkg_version in r["result"], r
        href = r["result"][pkg_version]["+links"][0]["href"]
        data = b""
        disconnect_len = (
            disconnect_at_length
            if disconnect_at_length > 0
            else len(content) + disconnect_at_length
        )
        with httpx.stream("get", href) as r:
            stream = r.iter_bytes(1024)
            while True:
                streaming_data = next(stream)
                data += streaming_data
                if len(data) >= disconnect_len:
                    break
    
        assert data != content
    
        pkg_file = files_path.joinpath(
            "root", "mirror", "+f", digest[:3], digest[3:16], pkgzip
        )
        # this is sometimes delayed a bit, so we check for a while
        for _i in range(50):
            if pkg_file.exists():
                break
            sleep(0.1)
>       assert pkg_file.exists()
E       AssertionError: assert False
E        +  where False = exists()
E        +    where exists = PosixPath('/tmp/test-szj_5id3-server-directory/primary/+files/root/mirror/+f/77d/640a11ff574a8/pkg1-1.0.zip').exists

test_devpi_server/test_streaming.py:200: AssertionError
```

I am really troubled by caching large package like pytorch, test this by download pytorch in browser, it does not work, I guess that you lost ownship of control flow once call `yield`, and must handle control flow in the same level of `entry` object.

fschulze:
> I've written a new version [YAGregor@4bbe5f3](https://github.com/YAGregor/devpi/commit/4bbe5f33f8f337b6f6a3e7b72397c74e6d4127e0) without catch GeneratorExit and copy paste, the key idea is

Is there any difference in regard to garbage collection between using ``finally:`` and catching ``GeneratorExit``? IOW, is the ``finally:`` branch not executed during garbage collection? If not, then your implementation just hides the handling of ``GeneratorExit`` instead of doing it explicitly.

fschulze:
I just tried it with a 500MB release of ``torch`` and it stops downloading after 5 seconds in the background with ``Client disconnected while serving ...``. So yeah, looks like control is lost. I fear this can only be fixed with either a separate thread, or using asyncio similar to what is done when downloading project data for mirrors/remotes. Problem is how to do that in a way which keeps the streaming going.

YAGregor:
> Is there any difference in regard to garbage collection between using `finally:` and catching `GeneratorExit`?

not really, the key point is let GeneratorExit raise from `file_streamer` to `iter_cache_remote_file` and drive it execute, I do that because I want to follow the document say never catch GeneratorExit, maybe reraise is another option.



> IOW, is the `finally:` branch not executed during garbage collection?

It will, but I think it's generator's caller's responsibility to close generator(end of for in also close a generator), but gc will close again.

a simple test
```python
def _iter():
    for i in range(10):
        try:
            print(i)
            yield i
        except GeneratorExit:
            print("generator close")
            raise


if __name__ == '__main__':
    for i in _iter():
        if i == 3:
            break
    print(">>>> end iter 1")

    iter2 = _iter()
    next(iter2)
    next(iter2)
    iter2.close()
    iter2.close()

    print(">>> end iter 2")

    iter3 = _iter()
```

```
0
1
2
3
generator close
>>>> end iter 1
0
1
generator close
>>> end iter 2
```

YAGregor:
> . Problem is how to do that in a way which keeps the streaming going.

Yes, in my opinion, the "correct way" should be manage all download process task in a [task group](https://docs.python.org/3/library/asyncio-task.html#asyncio.TaskGroup) by structured concurrency and send bytes to responses by message passing, there should be a global download task manager, but not let waitress server execute the download task, it's only the view layer anyway. But that will be a big change, currently I am using my patch, because I can't wait to use devpi to cache packages, especially large package like `torch`.

PR #1123, opened 2026-04-17
Add q, total, cached, limit, offset query parameters to GET /<user>/<index>

### LINKED ISSUE #1122: GET /<user>/<index>: add q, cached, total, limit, offset query parameters
## Problem

`GET /<user>/<index>` always returns the full project list with no filtering or pagination.
For large mirror indexes like PyPI this is ~780k projects / 17 MB per request.

This makes it impossible to build responsive UIs with lazy loading or search-as-you-type
without downloading the full list upfront on every page load.

## Proposal

Extend `GET /<user>/<index>` with opt-in query parameters. Without parameters the endpoint
behaves exactly as today — fully backwards compatible.

### Filtering

**`?q=<substring>`** — filter project names by substring match:
GET /root/pypi?q=flask

**`?cached=1`** — include a `cached` list in the response: projects that have locally
cached simplelinks (i.e. have been downloaded/accessed through this mirror):
GET /root/pypi?cached=1

### Pagination

**`?limit=<n>&offset=<n>`** — return a slice of the project list:
GET /root/pypi?limit=100&offset=0

When `limit` is present, the response includes `offset` and `limit` fields.

### Count

**`?total=1`** — include total count of (filtered) projects without fetching all of them:
GET /root/pypi?total=1&limit=100&offset=0

### Combined example

GET /root/pypi?q=flask&cached=1&total=1&limit=20&offset=0

{
  "result": {
    "type": "mirror",
    "projects": ["flask", "flask-login", "flask-wtf", ...],
    "cached": ["flask", "flask-wtf"],
    "total": 847,
    "offset": 0,
    "limit": 20
  }
}

## Use case

Building a dashboard UI ([devpi-admin](https://github.com/pavelrevak/devpi-admin)) that shows cached packages with search and
infinite scroll — without downloading 17 MB on every page load.

## Notes

- `?cached=1` uses existing `is_project_cached()` (DB lookup via keyfs, no filesystem scan)
- All parameters compose freely
- Mirror-only parameters (`cached`) are silently ignored on non-mirror indexes
- Parameters apply in order: q filter → total count → cached list → pagination slice

## Implementation

I'm working on a PR for this. Initial implementation will be ready for review soon.

pavelrevak:
## Summary

Extends `GET /<user>/<index>` with opt-in query parameters for filtering,
counting and paginating the project list. Without parameters the endpoint
behaves exactly as today — fully backwards compatible.

- `?q=<substring>` — filter project names by substring
- `?total=1` — include total count of (filtered) projects
- `?cached=1` — include list of locally cached projects (mirror indexes only)
- `?limit=<n>&offset=<n>` — paginate the project list

All parameters compose freely:
GET /root/pypi?q=flask&cached=1&total=1&limit=20&offset=0

## Motivation

Building a dashboard UI over the devpi API that needs search-as-you-type and
infinite scroll over the project list — without downloading 17 MB on every
page load.

## Notes

- `?cached=1` uses existing `is_project_cached()` (DB lookup via keyfs, no filesystem scan)
- `?cached=1` is silently ignored on non-mirror indexes
- Parameters apply in order: q filter → total count → cached list → pagination slice

Closes #1122

### DIFF
--- a/doc/curl.rst
+++ b/doc/curl.rst
@@ -56,12 +56,73 @@ Looking at list of projects in ``alice/dev`` index::
      $ curl -H "Accept: application/json" -s \
            -X GET http://localhost:3141/alice/dev/
      {
-       "status": 200, 
-       "type": "list:projectconfig", 
+       "status": 200,
+       "type": "list:projectconfig",
        "result": [
          "example"
        ]
      }
+
+The project list supports several optional query parameters:
+
+``?q=<substring>``
+    Filter project names by substring match::
+
+        $ curl -H "Accept: application/json" -s \
+              http://localhost:3141/root/pypi?q=flask
+        {
+          "result": {"projects": ["flask", "flask-login", "flask-wtf"]}
+        }
+
+``?total=1``
+    Include the total count of (filtered) projects in the response::
+
+        $ curl -H "Accept: application/json" -s \
+              "http://localhost:3141/root/pypi?q=flask&total=1"
+        {
+          "result": {"projects": ["flask", "flask-login"], "total": 2}
+        }
+
+``?cached=1``
+    Include a ``cached`` list — projects that have locally cached metadata
+    (mirror indexes only)::
+
+        $ curl -H "Accept: application/json" -s \
+              http://localhost:3141/root/pypi?cached=1
+        {
+          "result": {
+            "projects": ["flask", "django", "requests"],
+            "cached": ["flask"]
+          }
+        }
+
+``?limit=<n>&offset=<n>``
+    Paginate the project list. When ``limit`` is present, ``offset`` and
+    ``limit`` are included in the response. ``offset`` defaults to 0::
+
+        $ curl -H "Accept: application/json" -s \
+              "http://localhost:3141/root/pypi?limit=2&offset=0"
+        {
+          "result": {
+            "projects": ["aiohttp", "attrs"],
+            "limit": 2,
+            "offset": 0
+          }
+        }
+
+All parameters can be combined freely::
+
+    $ curl -H "Accept: application/json" -s \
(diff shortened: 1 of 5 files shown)


fschulze:
Whoa, I like your enthusiasm, but lets slow down a little with creating PRs before discussing a bit first.

The existing ``is_project_cached`` method is badly named. For ``get_cached_projects`` I would instead suggest either ``get_downloaded_projects_perstage`` or ``get_local_projects_perstage``, I prefer the latter as it could also be added to the regular index type and have proper meaning by just returning the project list. I also think you only need to call ``_stale_list_projects_perstage`` and not ``key_projsimplelinks(project).exists()`` which is pretty slow if many projects are downloaded, or did you get unexpected results without the check? In 7.x this will overall be more efficient and can be done with one database query.

For the REST API:
1. I think we should add a ``type=projectlist`` and return that as the type. I need that for 7.x as well. The "result" would be a list of dictionaries with at least ``name`` and ``normalized_name``, maybe more later.
2. ``offset`` is an inefficient pattern: https://botneve.com/api-design/api-pagination-patterns/ I would suggest an opaque cursor instead which internally can use the last returned name as the start for the next page, or the "keyset" variant from the article which is the same but visible in the API instead of opaque.
3. instead of ``cached`` I suggest either ``downloaded_only`` or ``local_only``, see above comments for ``get_cached_projects``.
4. ``q`` is too vague IMO and ``substring`` is slow if a database is used. I suggest ``startswith`` or something like that instead which can efficiently filter directly in the database with 7.x.
5. ``total`` is also inefficient if it applies to the number of results in the query, if it is always the total number of projects, regardless of filtering, then it should be fine and I can add an efficient function for that in 7.x.

pavelrevak:
Thanks for feedback, we can continue in discussion in #1122
for now I leaving this PR

Issue #1018, opened 2023-12-06
Support PEP 658 / PEP 714


vit-zikmund:
Hello, this is not a bug, rather a very humble request for enhancement :)

Quite a lot of python distribution tools start adopting [PEP 658](https://peps.python.org/pep-0658/) (and it's ugly little sibling [PEP 714](https://peps.python.org/pep-0714/)) which tackle distribution of a package's metadata in a separately downloadable file. This helps a great deal with dependency locking tools like `poetry` or `pip-compile`, as such step doesn't need to download the full package in order to get all the transitive dependencies. This is currently very annoying and time-consuming procedure, when big packages like `pytorch` are involved.

I've searched the sources here on github and didn't find any trace that this would be supported by `devpi`. As no one seems to ask yet about the author's take on these, here it goes :wink: 

I've [discovered](https://github.com/devpi/devpi/issues/986#issuecomment-1659590143) `devpi` already got support for [PEP 691](https://peps.python.org/pep-0691/), which as its own section for PEP 658 values, so these together might make a nice tight pack :smile:

fschulze:
You can try ``devpi getjson https://.../user/index/project/version`` to see what we have in the database.

vit-zikmund:
To sum up the specs, the core thing is:
1. extracting the wheel's `<package>-<version>.dist-info/METADATA` file (`sdist`s are technically also included in the spec, but not even supported by PyPI)
1. computing its sha256 checksum in the hex form
1. putting the file at the same path as the `<package-dist-name>.whl` as `<package-dist-name>.whl.metadata`
1. exposing its existence in the package listing with various names (lol, but yes) per the appropriate PEP, always containing the hash in some form or the other :laughing::
   - PEP 503 (simple HTTP API) - as an HTML attribute of the `<a>` tag holding the `.whl` download link:
      - [PEP 714](https://peps.python.org/pep-0714/#servers): `data-core-metadata="sha256=aea654168069f40d75ac2e4291a8267c720ebc1dcc6038259469290a67e8be14"`
      - [PEP 658](https://peps.python.org/pep-0658/#specification): `data-dist-info-metadata="sha256=aea654168069f40d75ac2e4291a8267c720ebc1dcc6038259469290a67e8be14"`
   - PEP 691 (JSON-based Simple API)
     - [PEP 714](https://peps.python.org/pep-0714/#servers):
        ```json
        "core-metadata": {
[...]

PR #1126, opened 2026-06-30
Fix core-metadata extraction for dotted package names. Refs #1018

changchinlan:
Requesting .whl.metadata for a cached wheel whose distribution name contains dots (e.g. jaraco.classes, ruamel.yaml) returns 500.

The dist-info directory inside the wheel keeps the original name verbatim (jaraco.classes-3.4.0.dist-info/), but the code reconstructs it from the normalized project name.

https://github.com/devpi/devpi/blob/d35f2b035ab386d9e915a1e7f12c4405a0cbab03/server/devpi_server/views.py#L1634-L1636

This PR fix the issue by find the single `*.dist-info/METADATA` entry in the zip namelist instead of guessing the name.
PEP 427 guarantees one dist-info directory per wheel.

For example, expected metadata url for jaraco.classes is:
```
https://files.pythonhosted.org/packages/7f/66/b15ce62552d84bbfcec9a4873ab79d993a1dd4edb922cbfccae192bd5b5f/jaraco.classes-3.4.0-py3-none-any.whl.metadata
```
which contains the dot in the filename.

### DIFF
--- a/server/devpi_server/views.py
+++ b/server/devpi_server/views.py
@@ -1631,11 +1631,18 @@ class PyPIView:
             return apireturn(502, e.args[0])
 
         if is_metadata:
-            metadata_filename = (
-                f"{entry.project.replace('-', '_')}-{entry.version}.dist-info/METADATA"
-            )
             with entry.file_open_read() as f, ZipFile(f) as zf:
-                wheel_metadata_contents = zf.read(metadata_filename)
+                # find the single .dist-info/METADATA in the wheel,
+                # as the dirname may not match the normalized project name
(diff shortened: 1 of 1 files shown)


fschulze:
Thanks for the PR, but I already fixed that yesterday in https://github.com/fschulze/devpi/commit/34c34fbb6f3cf6df94e9de1c0461b7237a138a03 and https://github.com/fschulze/devpi/commit/490ee822afd8ed280862124c28583f388d8a7c95 on my branch https://github.com/fschulze/devpi/tree/server6. I'm planning to release that later today.

